PR #7, opened 2018-09-26
Update version in readme

cjohansen:
The quickstart example in the Readme leads to a stack trace with 173. Update to 176 fixes it.


plexus:
Thanks for pointing this out, seems master didn't get pushed after the last release. See https://github.com/lambdaisland/kaocha/commit/7d3e303a014bd401095dc8a5e4c6855cf0abcee4 , this update happens automatically as part of the release process.

I'll push it out as soon as I'm home.

cjohansen:
👍

PR #13, opened 2018-11-06
Cloverage

plexus:
This integrates Kaocha with Cloverage through a plugin. This will be pulled out
into its own repo so as not to impose the extra dependency on people, but I'm
committing it here first to try it out on CI.

This introduces an extra `:main` hook that plugins can use to have their own
top-level behavior. They can shortcut any remaining activity by throwing
:kaocha/early-exit.

The plugin makes most Cloverage options accessible through both command line
arguments and tests.edn. Test and source paths are taken from test suite
configurations.

Note that apparently using a variable names "var" causes problems with


plexus:
Keeping this in the main repo for now, as juggling these mini repos is already getting tedious. People won't pay for it in extra load time unless they enable the plugin. This will get split out later as `lambdaisland/kaocha-cloverage`.

Released in `0.0-266`

PR #38, opened 2019-01-18
Add a pre-report hook

magnars:
This allows plugins to inspect and change test results just before they are
passed to the reporter.

Notes: 

- I had to do some wrangling to get the tests to run at all, ref the changes in monkey-patch.
- I am a little unsure of my change in `src/kaocha/plugin/hooks.clj` - I'm not sure if the hooks will be on the results.
- In the documentation I called the parameter into pre-report `results` - Is that a good name?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Added
 
+- Added a `pre-report` hook. This allows plugins to inspect and change test
+  events just before they are passed to the reporter.
+
 ## Fixed
 
 ## Changed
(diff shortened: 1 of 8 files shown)


magnars:
I just realized there must be other places in need of a pre-report hook. This only fixes clojure.test, right? Or does Midje fall back to this? cljs.test? I guess you must normalize the data structure between these somewhere, so that plugins don't need to consider different formats for different test libs?

plexus:
This is already looking really great! Thanks for thinking of the CHANGELOG and docs. There are two changes I think that still need to go in, one in the docs where it should say `pre-report` instead of `pre-result`, and one in the hooks plugin where it should say `test-plan` instead of `result`. Do you still want to make these or do I merge and fix it on master? (also fine by me)

magnars:
I'm happy to fix it in the PR. :-)

magnars:
All fixed, I hope. :)

plexus:
Thanks a lot!

thankyou

PR #41, opened 2019-01-20
Switch README links from cljdoc.xyz to cljdoc.org

danielcompton:


plexus:
Thanks! This toc is actually generated, I'll fix the script on master (it's in `bin/generate_toc`)

PR #43, opened 2019-01-21
Remove dependency on clojure.java-time

magnars:
It seems to be used for only one function that is just as easily implemented
using java.time directly.

I noticed this because I got an error when trying to evaluate kaocha in the repl. It had something to do with clojure.java-time. Since it was so easy to remove, I didn't spend much time figuring out why it failed.


plexus:
Good catch, always good to keep the dependencies in check a bit!

PR #48, opened 2019-01-23
Run all tests when --focus-meta doesn't match

### LINKED ISSUE #40: Suggestion: Run all tests if no tests match --focus-meta
I like to run all the tests in a terminal with `--watch`. Sometimes I'd like to focus on just one namespace or a few tests, while I work on them. Afterwards I go back to running all the tests again.

The way `--focus-meta` works now, I have to restart Kaocha twice to do this. How about running all the tests when there are no matching tests? It might have to be opt-in, or come with a warning, if you are concerned about confusion.

plexus:
Ignore :focus-meta tags that don't match any tests. Print a warning.

@magnars I know you would prefer not to have the warnings, but I'll open a separate ticket for making the warnings configurable.

Closes #40

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,9 @@
 
 - Added a `pre-report` hook. This allows plugins to inspect and change test
   events just before they are passed to the reporter.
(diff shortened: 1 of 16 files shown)


magnars:
Nice! Thank you. :-)

PR #54, opened 2019-01-24
A boatload of improvements to --watch

### LINKED ISSUE #16: Watch feature should ignore temporary files
It would be good if there was a way to specify patterns of filenames to ignore from the watch feature (e.g. `.#*` files created by emacs) so that tests are not needlessly reevaluated every time a temporary file changes.

### LINKED ISSUE #17: Watch mode improvements
There are several things regarding `--watch` mode that could be improved.

- [x]  A configurable include/exclude mechanism, seeded from `.gitignore` #16 
- [x]  Watch changes to `tests.edn` (related to 1., but also requires using a new test config)
- [x]  Drop `core.async`. It's the only place in the code where it's used, and a simple Java blocking queue would work just as well. This would help improve start up time.
- [x]  Don't build when a source namespace has load errors. Kaocha turns errors during the load step into test failures, so that one broken test ns doesn't stop the build from running, but errors in source namespaces should be treated as a broken precondition and fixed first.

### LINKED ISSUE #39: Suggestion: While watching, re-run all tests on first SIGINT
Sometimes changes are not detected (a git rebase, changes in non-clj files). Sometimes Kaocha is waiting for a single test to turn green, but you would like to run all the tests.

My suggestion is a SIGINT (Ctrl-C) would make Kaocha go:

```
Re-running all tests. Interrupt a second time to quit.
```

Then wait for two seconds for a second SIGINT before running all the tests.

I'd be happy to contribute something like this if it sounds reasonable.

plexus:
I've tried to address as many of the concerns and issues with --watch as
possible. In the process I've cleaned up the code quite a bit, but it's also
gotten quite a bit more complex, and it's still basically untested because this
stuff is a pain to test, but I've tried to dogfood it as much as possible.

The big issue is that is fixed is that load errors are handled much better. Now
if a namespace fails to load you'll get to see the relevant exception and trace,
and the whole thing is reported as if it's a failing test run. That way plugins
and reporters like the notifier plugin can also report on it. We no longer try
to do an actual run in this case, all tests are skipped.

Pressing enter will re-run all tests (not just previously failing tests).

You can set patterns for files to ignore in `tests.edn` (:kaocha.watch/ignore
[...]). These follow the glob spec of java's PatchMatcher. The original idea was
to source these from .gitignore, but .gitignore uses a pretty specific format,
so we'd have to parse and convert that, that's something for later.

Finally `tests.edn` will also be watched, and it gets reloaded on every run, so
you can change your config and add plugins on the fly.

Closes #16 
Closes #17 
Closes #39

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,10 +7,15 @@
 - Added a `:kaocha.plugin/notifier` plugin that pops up desktop notifications
   when a test run passes or fails.
 - Add the `wrap-run` hook to the hooks plugin.
+- Watch mode: re-run all tests by pressing "enter"
+- Watch mode: watch `tests.edn` for changes
+- Watch mode: ignore certain files with `:kaocha.watch/ignore [".*" ,,,]`
 
 ## Fixed
 
 - Preserve changes to the config made in a `pre-load` hook
+- Watch mode: if a namespace fails to load then report it clearly and fail the
+  test run, skipping any remaining tests.
 
 ## Changed
 
--- a/bin/generate_toc
+++ b/bin/generate_toc
@@ -4,7 +4,7 @@ def uslug(s)
   s.downcase.gsub(/[^\p{L}\p{N}]+/, ' ').gsub(/[\p{Z}\s]+/, '-')
 end
 
-Dir['doc/*.md'].sort.each do |f|
+Dir['doc/**/*.md'].sort.each do |f|
   title = File.read(f).each_line.first.sub(/^#+/, '').strip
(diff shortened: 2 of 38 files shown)


plexus:
Not merging this yet, I want to try to get some semblance of unit testing on this, and document some of the new behavior.

magnars:
Wow, is it christmas in january? 😄

plexus:
It took a bit of figuring out and reshuffling things, but there's a test now that hits most of the `kaocha.watch` code paths. I also went through the docs to give them a bit of a refresher, and implemented the tests that were marked as pending. I think this is good to go!

magnars:
This is an impressive number of improvements! Looking forward to using it. ❤️

PR #63, opened 2019-01-30
Stop the watcher from reacting to directory events

cjohansen:
On OSX changing files will often also emit events on the containing directory,
causing excessive test runs while in --watch mode. This manifests as duplicate
test runs, but is especially visible when ignoring certain file patterns.

This commit adds a check for directories and ignores those events.


plexus:
Thank you!

PR #68, opened 2019-01-31
Remove binding plugin, report load errors

### LINKED ISSUE #27: Make a distinction between load errors in the test file being loaded vs. in other namespaces.
If an exception occurs while Kaocha loads a test namespace, then a warning is printed, the `:kaocha.type/namespace` test is marked as failed, and loading/running continues.

This is so that when you have a typo which prevents a test namespace from loading, this doesn't stop the world. Instead it just runs the rest of the tests, and reports a failure pointing at the syntax error.

The problem with this approach is that loading a namespace could fail because a dependent namespace has issues. In the degenerate case a source namespace that is used indirectly in every test fails to load, causing a failure to be reported for every test namespace. This is especially easy to do (and annoying to encounter) in `--watch` mode.

Instead when loading a test namespace we should first evaluate its `ns` form. If that causes issues then we hard abort the test run. Only after the `ns` form has loaded do we load the full namespace, and turn load errors into test failures.

### LINKED ISSUE #59: Make kaocha.stacktrace/*stacktrace-filters* configurable?
In case I want to fine-tune them to my project?

plexus:
The binding plugin has been removed in favor of making the functionality
built-in. This way we can make sure bindings are installed during the complete
Kaocha process, including the load step.

Errors during the load step now cause the test run to fail. They are still
reported as if a test run happened, so plugins can pick up on them.

This reflects a change in my thinking. Originally my goal was to gracefully deal with load errors in tests, so that the rest of the tests still ran and provided good feedback, but because Kaocha made no distinction between errors in test namespaces and source namespaces this actually ended up really getting in the way, and hiding/obfuscating issues.

Rather than adding more logic to differentiate between test and source namespaces, I decided to keep things simple and report on load (syntax) errors loud and clear, and then bail out of doing anything else. If you have syntax errors those should really be fixed before anything else, since there's no telling how reliable other test results are under such circumstances.

Closes #59 
Closes #27


plexus:
This has been on master and released for quite some time, not sure why there's still an open PR.

PR #70, opened 2019-02-20
Suppressing warnings

magnars:
Hi @plexus !

Like we discussed in #40, I started looking at configuration options to suppress warnings. This is not a complete Pull Request, but I could use some feedback.

- Suppressing the `:focus-meta-not-found` warning works like this. 👍  

However, I was planning on adding flags for the other warnings as well:

- Suppressing the `:test-path-does-not-exist` warning only works if you add the configuration to the test suite, and not to the overall configuration map.

That's because the `testable` being passed in does not contain the configuration. How do you normally handle these cases? 

- Suppressing the `:deprecated-config-literal` seems to me like needless busywork. If you're able to add the suppressing flag to the config, just change to `#kaocha/v1` already. 😄 

Your thoughts on this are most welcome. 

Cheers!

### DIFF
--- a/src/kaocha/plugin/filter.clj
+++ b/src/kaocha/plugin/filter.clj
@@ -45,8 +45,9 @@
                          (testable/test-seq testable))
         focus-meta (set focus-meta)
         unused     (set/difference focus-meta used-meta)]
-    (doseq [u unused]
-      (output/warn "No tests found with metadata key " u ". Ignoring --focus-meta " u "."))
+    (when (:focus-meta-not-found (:kaocha/warnings testable) true)
+      (doseq [u unused]
+        (output/warn "No tests found with metadata key " u ". Ignoring --focus-meta " u ".")))
     (set/difference focus-meta unused)))
 
 (defn filter-testable [testable opts]
--- a/src/kaocha/testable.clj
+++ b/src/kaocha/testable.clj
@@ -63,7 +63,8 @@
   [testable]
   (load-type+validate testable)
   (doseq [path (:kaocha/test-paths testable)]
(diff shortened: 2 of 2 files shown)


plexus:
This looks like a good start. One thing to note is that this way things are configured on the test suite level, I would have rather expected them at the top level (with a possible test-suite level override).

magnars:
No, this is at the top level. That's why `:focus-meta-not-found` works like that, while `:test-path-does-not-exist` doesn't. 

> * Suppressing the `:test-path-does-not-exist` warning only works if you add the configuration to the test suite, and not to the overall configuration map.

-> this is me trying to ask how you normally pass down top level configuration to the test suite level so it is available there. :)

plexus:
> this is me trying to ask how you normally pass down top level configuration to the test suite level so it is available there. :)

In this case the processing is happening in a post-load hook, which receives the test-plan (and the test-plan is really just the config with more stuff added), so you could pass that all around until where you need it.

Or you can reference `kaocha.testable/*test-plan*` directly :) that should be bound at this point (I think).

alysbrooks:
@magnars I'm looking at older issues/PRs and cleaning them up. Are you interested in picking this up again?

magnars:
No, I wasn't able to complete it in a reasonable time. I've instead resorted to hiding the warnings in the tooling outside of kaocha (showing only number of warnings). Not perfect, but pragmatic for me.

alysbrooks:
@magnars Glad you found something that works, at least.

For you or anyone else interested, we would welcome work on this PR/a new PR using it as a basis, but I'm closing this for now.

Issue #60, opened 2019-01-26
Default ns-pattern "-test$" is confusing


tonsky:
Following README, it’s not clear why after everything is set up no tests have been found. Maybe change default ns-pattern value to `".*"`, with option to rewrite? Otherwise you might just get stuck, with no idea why it’s not working and how to continue. I spent a great deal of time figuring if it’s a classpath problem or what

plexus:
Related: #34

plexus:
I'm ok with changing this. I would generally recommend to stick to the Clojure convention of naming your test namespaces `something-test`, and to be selective so Kaocha doesn't load more than it needs to, but the current default is confusing. Better to load everything by default, but show in examples how to narrow it down.

fdserr:
> I would generally recommend to stick to the Clojure convention of naming your test namespaces `something-test`...

Departing from the Java convention, `clojure.test` 1.10 is more friendly with embedded tests, see [API](https://clojure.github.io/clojure/clojure.test-api.html) sections "DEFINING TESTS" and "OMITTING TESTS FROM PRODUCTION CODE". AFAIK test code is already being suppressed in ClojureScript under optimised compilation (conditions may apply). 

Indeed, not all tests should be embedded within code, but (eg. in Rust) it's really convenient to have unit tests close to the object under test.

Been bitten too, but a fix would also make embedding tests more straightforward.

alysbrooks:
I was looking over issues with the "first issue only" tag, and I'm not sure we need to do this any more. Thanks to #72, the quickstart is much clearer, (and I've since made that change to the docs) and #34 will help too. 

If we still go ahead and do it, I think we should introduce a new `#kaocha/v2` reader, so it doesn't affect existing configs that haven't specified `ns-pattern`.

plexus:
Agreed that this would be a good candidate for a `#kaocha/v2`, I've started a ticket to track that: #183.

Issue #34, opened 2019-01-18
Provide a warning when no tests are found


plexus:
This usually has to do with not configuring `test-paths` or `ns-patterns` correctly, and so the warning should hint at that.

plexus:
The latest version will now print a warning when `--focus` uses a test id that does not exist.

plexus:
We have this code in `kaocha.api`

```
                (when (empty? (:kaocha.test-plan/tests test-plan))
                  (output/warn (str "No tests were found, make sure :test-paths and "
                                    ":ns-patterns are configured correctly in tests.edn."))
                  (throw+ {:kaocha/early-exit 0}))
```

but that doesn't seem right. `:kaocha.test-plan/tests` at the top level will return all configured test suites, so this will always return true (if there is no configured test suite you get the default `:unit` test suite).

Instead we should use `kaocha.testable/test-seq`, this recursively finds all testables, ignoring the ones that marked as to be skipped. We can use that to find any "leaf" tests (we have three types of testables, suites, groups, and leafs. The first two only "group" other tests, only leaf tests contain actual test code). This can be checked with `kaocha.hierarchy/leaf?`.

So the condition should become something like this

``` clojure
(when-not (some hierarchy/leaf? (testable/test-seq test-plan))
   ... print warning and exit ...)
```

PR #72, opened 2019-02-23
Improve quick start

fdserr:
#60 "Default ns-pattern "-test$" is confusing"
#34 "Provide a warning when no tests are found"

1. Kaocha seems to be running okay out of the box without the need for a `test.edn` config file, provided some conventions are being followed. Make this the most straightforward option to get started. 

2. While the defaults make sense, a catch-all configuration example would help to get up and running quickly on existing projects, with a due statement that it is suboptimal.

3. When no tests are found, a warning should be printed, with the most obvious corrective actions briefly mentioned.

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,17 +145,21 @@ of the project to configure actual test and source paths, and optionally set a
 reporter or load plugins (cf. Configuration in the
 [documentation](https://cljdoc.org/d/lambdaisland/kaocha/)).
 
-Example `test.edn`:
+
+Example of a catch-all `test.edn` config file (should run all
+tests found in `src/` and `/test`, in any namespace).
 ``` clojure
 #kaocha/v1
-{:tests [{:id :unit
-          :test-paths ["test/unit"]
-          :source-paths ["src-clj"]
-          :ns-patterns ["-test$"]}]
- ;; :reporter kaocha.report.progress/progress
- ;; :plugins [:profiling :notifier]
+#kaocha/v1
+{:tests [{:id          :unit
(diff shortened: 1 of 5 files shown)


fdserr:
(1) Edit README and Configuration doc.

fdserr:
(2) Would this be a good enough catch-all configuration example to get started?
```
;test.edn
#kaocha/v1
{:tests [{:id          :unit
          :test-paths  ["test/unit" "src/clj"] ;all paths containing tests
          :ns-patterns ["\\S+"]}]} ;all namespaces names
```

fdserr:
(3) Would this be a good enough warning:
`Warning: no tests were found. Check that :test-paths and :ns-patterns are set in your configuration (test.edn).`

plexus:
Thanks @fdserr , this is looking good.

> (2) Would this be a good enough catch-all configuration example to get started?

Looks good to me. The pattern could even simply be `".*"`

> (3) Would this be a good enough warning:

This implies these values are not set, whereas they could be set but incorrectly. How about

"No tests were found, make sure :test-paths and :ns-patterns are configured correctly in tests.edn."

?

fdserr:
(2) Updated README

Having an issue with the config not being picked up.
https://github.com/lambdaisland/kaocha/issues/77

fdserr:
```sh
$ kaocha
WARNING: No tests were found, make sure :test-paths and :ns-patterns are configured correctly in tests.edn.
[]
0 tests, 0 assertions, 0 failures.
```

plexus:
Nice improvements, thanks a lot!

PR #74, opened 2019-02-28
Check for libraries versions

AndreaCrotti:
This is a simple way to blow up more noisily when a library version doesn't satisfy the needs of Kaocha (the org.clojure/tools.cli example).

Is not plugged in anywhere atm but for example setting a version higher than the one need would give a similar error at the moment.

screen shot 2019-02-28 at 11 47 06

### DIFF
--- a/deps.edn
+++ b/deps.edn
@@ -13,7 +13,8 @@
   orchestra                    {:mvn/version "2018.12.06-2"}
   aero                         {:mvn/version "1.1.3"}
   progrock                     {:mvn/version "0.1.2"}
-  meta-merge                   {:mvn/version "1.0.0"}}
+  meta-merge                   {:mvn/version "1.0.0"}
+  andreacrotti/semver          {:mvn/version "0.2.2"}}
 
  :aliases
  {:test
--- a/src/kaocha/versions.clj
+++ b/src/kaocha/versions.clj
@@ -1,0 +1,23 @@
+(ns kaocha.versions
+  (:require [clojure.java.io :as io]
+            [semver.core :as semver])
+  (:import (java.util Properties)))
+
+(defn get-version
+  [dep]
+  (let [path (str "META-INF/maven/" (or (namespace dep) (name dep))
+                  "/" (name dep) "/pom.properties")
+        props (io/resource path)]
+    (when props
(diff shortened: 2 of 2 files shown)


danielcompton:
There's not a ton of code here, but you could consider using https://github.com/trptcolin/versioneer as well for getting the version of the projects, it has a little more handling for failure cases and other edge cases.

plexus:
Have we measured how much start up time this adds? I'm a bit weary of adding more libraries that need to get loaded on every run, for a check that only matters for a small set of users.

Since tools.cli is the only dependency version issue people have run into so far, I'd first look at how we can address this specific case, before introducing more generic mechanisms.

The "simplest thing that might just work" is to add a call to tools.cli that parses a `--[no-]foo` argument. From the result we can quickly see if the loaded version is affected. This has the added benefit that it's not dependent on Maven implementation details, e.g. it still works even when people pull tools.cli as a Git or local dependency.

Other options
- vendor the tools.cli code inside Kaocha
- push tools.cli 0.4.0 to clojars under a different artifact id and namespace

Apart from that I think it's also worthwhile to see which libraries are causing the old tools.cli to be pulled in, and to encourage those projects to upgrade. @AndreaCrotti @magnars  @stathissideris you all ran into this, do you remember which library caused the conflict?

magnars:
In our case, it was Datomic Pro. 

```
[org.clojure/tools.cli "0.4.1"] ;; must specify here override the old version in datomic-pro, for kaocha
[lambdaisland/kaocha "0.0-389"]
```

I'll throw in another possibility, that certainly has some trade-offs: https://github.com/benedekfazekas/mranderson

stathissideris:
For me, kaocha couldn't recognise the `--watch` flag when I had this in my deps:

`[org.clojure/tools.cli "0.3.5"]`

AndreaCrotti:
In my case @plexus we were just including an older version of tools.cli explicitly.
And to be fair I think most projects in my company at least would have the same problem.

I haven't measured the difference in startup time, but well if tools.cli is really the only library that can cause problems it's fine to just do the call you suggest.
But well I just think lots of people will have this problem (given how common is to include tools.cli), so giving a nicer error would ensure they don't waste too much time.

PR #78, opened 2019-03-05
Add a check to make sure clojure.tools.cli is up to date

plexus:
In particular check its capabilities, so that we're sure it can handle all our
use cases. The handling of --[no-]xxx flags was only added in 0.3.6.

It's a very specific and perhaps slightly hacky check, but it's also virtually free in terms of cycles spent and code added, so it seems like a good starting point.

@AndreaCrotti I hope this suffices for you. 

Closes #74

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Added
 
+- Added a check to make sure org.clojure/tools.cli is up to date
+
 ## Fixed
 
 ## Changed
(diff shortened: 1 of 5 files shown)


danielcompton:
I might amend the message to say:

```
(output/error "This version of org.clojure/tools.cli does not have all the capabilities that Kaocha needs ...
```

But it's a little bit tricky to write the message here because you're wanting to talk about where the dependency came from, which could be many places, so it's hard to be too specific.

AndreaCrotti:
Yes of course @plexus , I can't see any changes or commits in this PR but that would be fine for me.

plexus:
Released as `0.0-413`

PR #85, opened 2019-04-29
Make scripts more portable, replacing  `/bin/bash` with `/usr/bin/env bash`

jlesquembre:
The path `/bin/bash` doesn't exist in  some unix systems (like [NixOS](https://nixos.org/), which I use).
In general it's more portable to use `/usr/bin/env bash` instead of `/bin/bash`, see https://en.wikipedia.org/wiki/Shebang_(Unix)#Portability

It's just a suggestion, but I thought it would be better to offer the suggestion with a PR

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,7 +83,7 @@ Add a binstub called `bin/kaocha`
 
 ```
 mkdir -p bin
-echo '#!/bin/bash' > bin/kaocha
+echo '#!/usr/bin/env bash' > bin/kaocha
 echo 'clojure -A:test -m kaocha.runner "$@"' >> bin/kaocha
 chmod +x bin/kaocha
 ```
@@ -104,7 +104,7 @@ Add a binstub called `bin/kaocha`
 
 ``` shell
 mkdir -p bin
-echo '#!/bin/bash' > bin/kaocha
+echo '#!/usr/bin/env bash' > bin/kaocha
 echo 'lein kaocha "$@"' >> bin/kaocha
 chmod +x bin/kaocha
 ```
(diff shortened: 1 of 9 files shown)


danielcompton:
I'm having trouble finding reliable sources, but I do remember `/usr/bin/env bash` being slightly slower than calling `/bin/bash` directly. Not sure whether it's really a problem here, but thought it worth mentioning.

https://unix.stackexchange.com/a/206357

jlesquembre:
@danielcompton `/usr/bin/env bash` is slower, since looks for `bash` in your `PATH` and then calls it, while `/bin/bash` doesn't need to look in your `PATH`. But the difference is minimal, `env` is pretty fast (`/usr/bin/env bash` takes 6ms on my computer)

plexus:
@jlesquembre can you elaborate why you want this change? The scripts under bin/ are only relevant for people working on Kaocha itself, most of them are only relevant if you're releasing new versions. Are you planning to contribute more to Kaocha?

People who have their bash in a non-standard location presumably are aware of that and can adjust the installation instructions accordingly.

I'll go ahead and merge this, although I'm fairly torn on it. It trades compatibility for a small contingent of users for a (admittedly small) security risk and a (small) invocation overhead.

jlesquembre:
@plexus Honestly, the only reason to create this PR was because I just started to use Kaocha, and after copy/pasting the commands from the quickstart documentation, Kaocha binstub was returning an error on my system. I thought it will be nice to fix that for other people.

Instead of opening an issue, I thought it will be better to directly create a PR, to help you to make the changes in case you wanted to accept them. For completeness, I also updated the scripts, but my first idea was to send a PR for the documentation only. I don't have any concrete plans on contributing to Kaocha, but since I started to use it, if I see something were I can contribute, I'll do.

As I said, this PR is just a suggestion, if you don't like it, it's totally fine with me if you decide to revert the changes.

plexus:
Thanks for clarifying. I'll leave it like this, I don't think the downsides weigh up to the fact that it may help some people.

PR #91, opened 2019-05-20
Fix option docs for --no-capture-output

thobbs:
The correct option is `--no-capture-output`, not `--no-capture`


plexus:
Thank you!

PR #95, opened 2019-05-29
Automatic spec test check generation

### LINKED ISSUE #45: Support running clojure.spec.test.alpha/check
It might be good to consider how to support generative testing suites through clojure.spec and a kaocha wrapper over the `clojure.spec.test.alpha/check` runner (or possibly by providing an alternative implementation).

Here's a proposal as to how this might work:

Firstly there are a few behaviours of `check` that have implications that will need to be considered.

1. Calling the 0-arity `(check)` finds all loaded `fdef` specs and executes them.  The difficult thing here is that the `fdef`s and specs need to be loaded already for them to execute.  So we'd need to provide a mechanism to do this.  `fdef`s are named by the vars of the functions they check, so there is no way to provide a naming convention feature to find them.  

The easiest approach here is probably just to let the user provide a set of namespaces to load before we trigger `check` to locate and test the specs, through something like `:kaocha/ns-patterns "myapp.*"`.  Spec "tests" will likely need their own `:kaocha.testable/type` (I'd suggest `:kaocha.type/clojure.spec.test.alpha.check` to allow for alternative runners in the future).  It's also worth bearing in mind there may not be a need for the equivalent of `:kaocha/test-paths` as specs commonly exist in the application namespaces.  

2. The 1-arity `(check sym-or-syms)` allows you to provide either a single symbol representing an fdef to check, or a set of symbols to check.  This should be configurable via the `tests.edn` and command line.

3. `(check sym-or-syms opts)` takes the symbol/set as with the 1-arity but also takes a map of `test.check/quick-check` [options](https://clojure.github.io/test.check/clojure.test.check.html#var-quick-check) and `:gen` overrides.  These should both be configurable in the config, however only the `quick-check` options should be settable via the command line (it doesn't make much sense to supply `:gen`s via the commandline).  In particular these include: the `:seed` option which can likely be set to the same seed `kaocha` uses, and `:max-size` which is required for constraining the size of inputs and is very important to set if you want performant tests.  Additionally it is important that users can set specs additional `:num-tests` property here which specifies the number of random tests to generate for each `fdef`.

I think it may also be useful for users to be able to orchestrate a set of `check` tests where multiple `check` calls can be made with different settings for  each of the above options within a single suite (as with `:core-specs` suite below:

```clojure
#kaocha/v1
{:tests     ;; [1]
    [{:id    :core-specs
              :type :kaocha.type/clojure.spec.test.alpha.check
              :kaocha/ns-patterns "myapp.*"
              :source-paths ["src"]
              :checks [{;; :syms #{myapp.foo/bar myapp.foo/baz}  ;; omitting :syms means find all fdefs
                              :clojure.spec.test.check {:num-tests 100 :max-size 20 }
                              {:syms #{myapp.flib/flip myapp.flib/flop} 
                              :clojure.spec.test.check {:num-tests 10 :max-size 10 }
      {:id :slow-specs
            :type :kaocha.type/clojure.spec.test.alpha.check
            :kaocha/ns-patterns "myapp.*"
            :source-paths ["src"]
            :clojure.spec.test.check {:num-tests 1000 :max-size 50 }
            :checks [{:syms :all-fdefs}]]}]}

```

Note here that options can be specified at multiple levels, with the inner in `:checks` overriding those provided in the wrapping layer.  I think a reason to do this is to avoid complecting the suite with the sizes you need to supply to a test etc.  Some specs are much more sensitive to sizing parameters than others, and may for example not terminate without large heap settings.  

I think any sizing/num-tests options provided at the command line though would override everything within the config and apply to the whole suite.

Finally we would need to consider reporting.  `check` emits a lazy sequence of reports containing the following information 

```
:spec       the spec tested
:sym        optional symbol naming the var tested
:failure    optional test failure
::stc/ret   optional value returned by test.check/quick-check
```

These would then need to be summarised and printed in human readable form.  Additionally if any errors/exceptions occur we should report them nicely with context.  We should parse an `::s/failure` data if it's present  and provide a human readable explanation of the problem:

```
The value for :failure can be any exception. Exceptions thrown by
spec itself will have an ::s/failure value in ex-data:

:check-failed   at least one checked return did not conform
:no-args-spec   no :args spec provided
:no-fn          no fn provided
:no-fspec       no fspec provided
:no-gen         unable to generate :args
:instrument     invalid args detected by instrument
```

WhittlesJr:
Resolves #45.

Add automatic generative spec checking to kaocha. Previously, if you wanted to take advantage of clojure.spec's mighty generative testing capabilities, you had to write quite a bit of boilerplate. But I was sold on clojure.spec's notion of "free" generative testing, so why isn't it actually free? 

Up until now I've been using a variation of [this gist](https://gist.github.com/kennyjwilli/8bf30478b8a2762d2d09baabc17e2f10) to get what I wanted, but I knew we could do better. 

With this feature, all you need is kaocha. You don't need to write anything whatsoever to create tests for your specs (i.e. no more `defspec-test` from that gist); kaocha will discover and generate them automatically. Now, writing your spec *is* writing your test.  (I mean, to a point :P clojure.spec doesn't know if your function is actually doing what you want it to do, but this catches a lot of errors for free.) 

This is a big deal for me and my workflow, and I suspect that others can benefit from this as well.


src/kaocha/load.clj
```diff
@@ -21,8 +21,8 @@
              (filter (partial ns-match? ns-patterns)))
             test-paths))
 
-(defn load-test-namespaces [testable ns-testable-fn & [platform]]
-  (let [test-paths  (:kaocha/test-paths testable)
+(defn load-namespaces [testable paths-key ns-testable-fn & [platform]]
```
plexus: This function is also used in other libraries like `kaocha-midje`, so we should keep this signature intact. Instead I think there's a simple refactoring where `load-test-namespaces` pulls out `:kaocha/test-paths`, and then calls `load-namespaces` with the paths.

``` clojure
(defn load-namespaces [testable paths ns-testable-fn & [platform]]
  ,,,)

(defn load-test-namespaces [testable ns-testable-fn & [platform]]
  (load-namespaces testable (:kaocha/test-paths testable) ns-testable-fn platform))
```

src/kaocha/plugin/alpha/spec_check.clj
```diff
@@ -0,0 +1,41 @@
...
+  (-> {:kaocha.testable/type    :kaocha.type/clojure.spec.test.alpha.check
+       :kaocha.testable/id      :generative
+       :kaocha/ns-patterns      ns-patterns
+       :kaocha/source-paths     ["src"],
+       :kaocha.filter/skip-meta [:kaocha/skip]
+       ::type.stc/syms          :all-fdefs}
+      (merge (type.fdef/stc-opts config))))
+
+(defplugin kaocha.plugin.alpha/spec-check
+  (pre-load [config] (update config :kaocha/tests conj (testable config)))
```
plexus: It seems the idea here is that people invoke these tests through the plugin? i.e. instead of having an entry in `tests.edn` they enable the plugin?

I'd rather turn it around. I've been thinking for a while that test types should be allowed to register plugins, so you can have this plugin be automatically active when someone adds a test suite of type `:kaocha/clojure.spec.test.alpha.check`, so that people can then use the `--spec-check-num-tests` and `--spec-check-max-size`command line flags.

src/kaocha/plugin/alpha/spec_check.clj
```diff
@@ -0,0 +1,41 @@
...
+           :parse-fn #(Integer/parseInt %)]
+          [nil  "--max-size SIZE" "Maximum length of generated collections"
+           :parse-fn #(Integer/parseInt %)]
+          [nil  "--ns-pattern PATTERN" "Regex matching your project's namespaces"]))
+  (config [config]
+    (let [num-tests  (get-in config [:kaocha/cli-options :num-tests])
+          max-size   (get-in config [:kaocha/cli-options :max-size])
+          ns-pattern (get-in config [:kaocha/cli-options :ns-pattern])]
+      (assoc config
+             ::ns-patterns [ns-pattern]
```
plexus: You're reusing the existing `:kaocha/ns-patterns` key here, but on the config instead of on the testable. That's going to cause confusion. Maybe use `::stc/ns-patterns` instead?

src/kaocha/type/clojure/spec/alpha/check.clj
```diff
@@ -0,0 +1,32 @@
...
+            [kaocha.type.clojure.spec.alpha.ns :as type.spec.ns]))
+
+(alias 'stc 'clojure.spec.test.check)
+
+(defmethod testable/-load :kaocha.type/clojure.spec.test.alpha.check
+  [{::keys [syms] :as testable}]
+  (-> (if (= syms :all-fdefs)
+        (load/load-namespaces testable :kaocha/source-paths type.spec.ns/->testable)
+        (assoc testable :kaocha.test-plan/tests (type.fdef/load-testables syms)))
+      (testable/add-desc "clojure.spec.alpha.check")))
```
plexus: How come this testable type does not have a `-run` method?

src/kaocha/type/clojure/spec/alpha/fdef.clj
```diff
@@ -0,0 +1,87 @@
...
+
+(s/def :kaocha.fdef/name :kaocha.var/name)
+(s/def :kaocha.fdef/var :kaocha.var/var)
+(s/def :kaocha.fdef/sym symbol?)
+
+(s/def :kaocha.type/clojure.spec.alpha.fdef
+  (s/keys :req [:kaocha.testable/type
+                :kaocha.testable/id
+                :kaocha.fdef/name
+                :kaocha.fdef/var]))
```
plexus: There are a couple things missing here, like I said this part is the least well documented so that's my bad.

- `:kaocha.testable/wrap` this contains wrapping functions that need to be wrapped around your main test logic, so plugins like the output-capturing work
- `(hierarchy/derive! :kaocha.type/clojure.spec.alpha.fdef :kaocha.testable.type/leaf)` this test type never has children, so it's a leaf type. this is needed for some plugins and reporters.
- `(t/do-report ...)` I don't see any `:begin-...` and `:end-...` events for these test types. Does `stest/check` emit certain `clojure.test` events?

src/kaocha/type/clojure/spec/alpha/ns.clj
```diff
@@ -0,0 +1,29 @@
...
+            [kaocha.core-ext :as core]
+            [kaocha.hierarchy :as hierarchy]
+            [kaocha.testable :as testable]
+            [kaocha.type :as type]
+            [kaocha.type.clojure.spec.alpha.fdef :as type.fdef]
+            [kaocha.type.ns :as type.ns]))
+
+(def ->testable (partial type.ns/testable :kaocha.type/ns))
+
+(defmethod testable/-load :kaocha.type/ns-spec-fdefs [testable]
```
plexus: `ns-spec-fdefs` doesn't seem to be referenced anywhere?

plexus:
One thing that would be really handy is to add a cucumber test that shows how you expect this to be used in practice.

src/kaocha/type/clojure/spec/alpha/ns.clj
```diff
@@ -0,0 +1,29 @@
+(ns kaocha.type.clojure.spec.alpha.ns
+  (:require [clojure.spec.alpha :as s]
+            [clojure.spec.test.alpha :as stest]
+            [kaocha.core-ext :as core]
```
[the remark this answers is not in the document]
WhittlesJr: I'll move and rename `in-namespace?` as you suggested

src/kaocha/plugin/alpha/spec_check.clj
WhittlesJr: My goal is just to allow people to add some simplistic spec checking purely through the CLI. But as I have it written, you could also add things to `tests.edn` if you need more fine-grained control, per @RickMoynihan's proposal.

I personally don't like maintaining `tests.edn`s since I have quite a few projects, so I wanted a way to do this through just CLI. That's the only purpose of the plugin. Is there a better way to offer that functionality?

src/kaocha/type/clojure/spec/alpha/check.clj
WhittlesJr: Just haven't gotten that far yet. I think the implementation will be nearly the same as `:kaocha.type/clojure.test` so I probably want to reuse it outright, don't I?

src/kaocha/type/clojure/spec/alpha/fdef.clj
WhittlesJr: I probably should add the `:begin-` and `:end-` reports, I haven't gotten it all running yet but I probably will today!

src/kaocha/type/clojure/spec/alpha/ns.clj
WhittlesJr: Probably just a WIP issue. While we're on the topic, do you think this is an appropriate name for "all the clojure.spec fdefs in the namespace?"

src/kaocha/plugin/alpha/spec_check.clj
plexus: Isn't it easier to add 

```
#kaocha/v1
{:tests [{:type :kaocha.type/clojure.spec.test.alpha.check}]}
```

to the project once, rather than adding `--plugin kaocha.plugin.alpha/spec-check` to each invocation?

WhittlesJr: Not in my case, since I have a lot of projects, and I can just CTL+R to recall my previous test run invocation for any given project. (Or put it in a binstub somewhere on my system). That's my preferred workflow.

WhittlesJr: Er I guess since we have to specify ns-patterns it would be a different invocation per project, but it's still not necessarily something I'd want to manage in source control for each individual project. That may change, though, if some projects have really specific tuning requirements. But I like to have the option of CLI-only config.

WhittlesJr: (furthermore, I'd love to not require ns-patterns. I'd much prefer if we could figure that out automatically. I'll look into that.)

plexus: This does go against Kaocha's philosophy, one of the core ideas is that your test configuration lives with the project so that you can use the same invocation (`bin/kaocha`) everywhere.

WhittlesJr: Maybe when I get to a place where my kaocha workflow is more stable, I could do that. But right now it would be hard to maintain across all the projects while I figure out the workflow I want. I started out using `tests.edn` but ran into this problem, so I switched to CLI arguments. I love that your project supports both. 

Also, right now I'm the only developer on these projects, but when (if) other developers join me, it strikes me as something that each developer might want to run a little differently (different reporting preferences, different performance requirements, etc.). So they'd possibly want to have their own version of `tests.edn`, in which case it wouldn't be source controlled.

WhittlesJr:
OK, I have it working for my simple use case. I haven't tried more fine-grained test configuration in a `tests.edn`, but it should be supported pretty much exactly as the OP of #45 suggested. 

I added two extra options: ::stc/instrument? and ::stc/check-asserts?. These are useful for catching bugs that stest/check doesn't catch on its own. 

My test invocation is:

`clojure -Atest -m kaocha.runner --reporter kaocha.report/documentation --plugin kaocha.plugin.alpha/spec-test-check --num-tests 50 --max-size 2 --instrumentation --spec-asserts`

Obviously this still also needs documentation and unit test coverage, but oh man is this feature nice. Like I said in #45 , this is the holy grail of testing for me. It's this type of "free" test coverage that really sold me on clojure.spec in the first place. 

I'm excited to see your revisions and polish! Once you're done, I can add documentation and tests.

WhittlesJr:
Alright, I'm happy with the PR at this point, but if you want me to go in and add docstrings and tests I can do that too.

plexus:
If you could add at least one cucumber integration test that'd be great. They're fairly straightforward, just copy over what you need from the other tests. With that I think this can be merged, and we can continue on master or move towards pulling it into its own repo.

WhittlesJr:
Well, I wrote it, but I had to use --no-color for the text-matching to work. None of your other Cucumber tests seem to use that...

WhittlesJr:
I definitely want to come back and write more of these tests, but I won't have much time in the near term.

WhittlesJr:
Hmm, I'm also not sure why it fails in CircleCI... it passes on my system, and nothing in the output looks incorrect.

plexus:
I had a look, the problem with your integration test was that after swapping the merge of the defaults, the test was now getting the default value of `:kaocha/ns-patterns` as provided by the logic that loads `tests.edn`, so it was only looking for fdefs in namespaces that ended in `-test`.

Instead I introduced its own config key, `:kaocha.spec.test.check/ns-patterns`, so that they don't clash.

Let's see but I expect it will run now. 

Thanks for adding the test, I think this is good enough to go onto master now. One thing I was thinking is that we could shorten some of the configuration keys

```
:kaocha.type.clojure.spec.test.check/checks  ->  :kaocha.spec.test.check/checks
`:kaocha.type.clojure.spec.test.check/syms  ->  :kaocha.spec.test.check/checks
```

etc

Similarly perhaps `kaocha.type/clojure.spec.test.check` could just be `:kaocha.type/spec.test.check`.

plexus:
_now_ it should work

WhittlesJr:
Whoa, that blew my mind ... I did not realize you could push to my fork's branch! I don't mind but I don't remember [granting permission](https://help.github.com/en/articles/committing-changes-to-a-pull-request-branch-created-from-a-fork)... just for future reference, do you know how that happened?

WhittlesJr:
I agree with the renames as well, I'll do that

plexus:
Hah, good point. Really not sure what the default is there or when Github switches it on.

WhittlesJr:
Oh, speaking of abbreviation, I noticed that in fdef.clj some of the log reports use `:kaocha.stc/` as the namespace. Should we use that everywhere instead of `kaocha.spec.test.check`? Or should we expand `:kaocha.stc`? Or keep them different?

WhittlesJr:
I updated the keys with your suggestions, added a plugin test, and fixed a bug with the plugin. I also rebased onto master, which picked up some unit test failures from the --watch fix... should I have waited?

WhittlesJr:
Myea forget the rebase. I like having the commits chronologically interleaved with the comment history.

plexus:
I'll read the diff over once more to see if we missed anything but I'm guessing this is good to go. I'll merge this, and then make a new PR to fix the test failures and address the issue with --watch.

plexus:
> `:kaocha.stc` ... Should we use that everywhere instead of kaocha.spec.test.check

I'd say we standardize on `kaocha.spec.test.check`

PR #96, opened 2019-06-02
Add type hints to eliminate reflection warnings

zharinov:
Hey folks, thank you for the Kaocha! 
I'm running with with `*warn-on-reflection*` set as `true` and you may found useful to have these type hints as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Added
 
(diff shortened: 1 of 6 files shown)


plexus:
Thanks, this is great! Could you also add a mention to the CHANGELOG? Thx!

plexus:
:ok_hand:

PR #108, opened 2019-07-26
Improve documentation on test identifiers

RickMoynihan:


doc/06_focusing_and_skipping.md
```diff
@@ -30,9 +35,9 @@ or namespace metadata, based on four command line flags and configuration keys.
 
 ## Matching
 
-### On id
+### by var
```
plexus: I would leave this title, since not all test types have vars (e.g. cucumber), but all tests have ids.

doc/06_focusing_and_skipping.md
```diff
@@ -9,14 +9,19 @@ For example:
 - Skip tests marked as slow
 - Focus on a test that previously failed
 
-You can skip tests, or focus on tests, either based on the test ID, or on test
+You can skip tests, or focus on tests, either based on a test `ID`, or on test
 or namespace metadata, based on four command line flags and configuration keys.
 
+Test `ID`s can be either a keyword corresponding to the `:tests` `:id` in 
+`tests.edn`, or a metadata flag, or a symbol identifying the test namespace or
```
plexus: > Test `ID`s can be ... a metadata flag

Let's not conflate these two. We shouldn't refer to a metadata flag as an "ID". It's not related to test ids, it's just another way to filter, and it's always a keyword.

plexus:
Thanks a lot!

PR #114, opened 2019-09-30
Correct plugin in keywords in README

RickMoynihan:


plexus:
Thanks, this was actually meant to work as advertised, i.e. unqualified implies `kaocha.plugin/...`. This works on the command line but not yet in tests.edn.

Issue #116, opened 2019-10-09
:no-gen documentation appears to be imprecise


imrekoszo:
I'm using the `kaocha.plugin.alpha/spec-test-check` plugin and found that in order to exclude an fdef-ed function from the automatic test run I have to decorate function definition itself with `^:no-gen` as opposed to the `fdef` itself as the documentation suggests.

However it is completely possible that I (and recently a colleague) misunderstood the documentation.

On the other hand if the documentation is indeed incorrect or ambiguous, you might consider reviewing my PR that attempts to clarify it.

PR #117, opened 2019-10-09
Correct usage documentation of :no-gen

imrekoszo:
Fixes https://github.com/lambdaisland/kaocha/issues/116


plexus:
I wasn't involved in writing this plugin. @WhittlesJr does this look good?

That said I think we should replace `:no-gen` with `:kaocha/no-gen`. This makes it clear to a person reading the code that this is kaocha-related metadata, and means we don't clash with other tools.

WhittlesJr:
I believe this is an appropriate change. I probably didn't realize the distinction because I never write my fdefs separately from my functions. (I use `defn-spec` from orchestra.) I also agree that :kaocha/no-gen is a good default value for `:kaocha.filter/skip-meta`.

plexus:
Thanks a lot @imrekoszo

imrekoszo:
Any time, and thanks for this great piece of sw!

PR #120, opened 2019-10-24
Fix documentation for progress reporter

opqdonut:
kaocha.report.progress/progress only shows the bar,
kaocha.report.progress/report also shows the actual failures, so it's
probably what people want

Could also consider renaming `progress/progress` -> `progress/progress*`, `progress/report` -> `progress/progress`.


plexus:
Thanks for catching this!

PR #125, opened 2019-11-26
Fix typo

emlyn:


plexus:
I'll merge this for now, but note that these files are generated from cucumber tests ([source](https://github.com/lambdaisland/kaocha/blob/master/test/features/clojure_test/assertions.feature)), so next time we generate them this backslash will come back. The reason it's there is that otherwise Cucumber will think the `When` is a special keyword.

This should really be fixed in the cucumber->markdown export code, which is an ad-hoc hack at this point. Might be easier to rephrase this paragraph so it doesn't start with `When`.

Thanks for the patch, welcome to Kaocha!

emlyn:
Thanks, really liking it so far!

PR #128, opened 2019-12-19
Introduce a meta-merge aero tag

### LINKED ISSUE #127: Support a meta-merge aero reader tag
It’s great that you can use aero in config… however one thing I’ve found (both in using aero with kaocha and in numerous projects using aero elsewhere).  Is that when trying to build profile cascades with aero; where you have a base config and a local.edn file of develper specific customisations, aero gets really messy and hard to DRY up.

e.g. I have this base test.edn config:

```clojure
#kaocha/v1
#merge [{:tests [{:id          :static-data
                  :test-paths  ["test"]
                  :source-paths ["src"]
                  :ns-patterns ["-test$"]
                  :skip-meta [:cleans-database]}
                 {:id         :mutating-data
                  :test-paths  ["test"]
                  :source-paths ["src"]
                  :ns-patterns ["-test$"]
                  :focus-meta [:cleans-database]}]
         :reporter kaocha.report/documentation
         :plugins [:kaocha.plugin/instrument-specs
                   :kaocha.plugin.some.required.project.specific-plugin
                   :kaocha.plugin/profiling
                   :kaocha.plugin/instrument-timings
                   :kaocha.plugin/print-invocations]}
        #include "tests.user.edn"]
```

If a user wants to add their own extra plugin, e.g. :kaocha.plugin/notifier they have to copy the whole set of plugins from the base config into their own config.

This can then be problematic when the base config changes without the user knowing, as they’re stomping the whole set of plugins, not just adding a new one.

For these scenarios I’ve found meta-merge across whole files of config to be a far superior approach.

I’ve not found a suitable solution to this with pure aero; however I thought I’d mention that on an internal project of ours which had this issue; we added meta-merge as a dependency, and 
exposed it through aero with a two-line aero/reader extension:

```
(defmethod aero/reader 'meta-merge [opts _tag value]
  (apply meta-merge value))
```

This one change then let replace almost all of aero with a single #meta-merge in place of #merge at the top, and aero #include

RickMoynihan:
Fixes #127 

I suspect the diff coverage fell beneath the threshold because I also moved a function down within the file.

There are no tests for this feature.  I can add some but not sure how best to test it without testing meta-merge.  Could maybe just test a file the meta-merge tag in it doesn't fail a call to read?

### DIFF
--- a/doc/03_configuration.md
+++ b/doc/03_configuration.md
@@ -53,6 +53,36 @@ used in `#kaocha/v1 {}`, rather than using fully qualified keywords.
 Configuration is read with [Aero](https://github.com/juxt/aero), meaning you
 have access to reader literals like `#env`, `#merge`, `#ref`, and `#include`.
 
+Additionally kaocha supports an extra `#k/meta-merge` reader tag, that
+works similarly to aero's `#merge` but will merge a whole subtree with
+the semantics of
+[meta-merge](https://github.com/weavejester/meta-merge).  This is
+particularly useful for configuring a cascade of includes that
+selectively augment and override a base configuration.  e.g. imagine a
+base tests.edn file with a `#k/meta-merge` and an `#include`:
+
+```
+#kaocha/v1
+#k/meta-merge [{:tests [{:id         :unit
+              :test-paths ["test/unit"]}
+             {:id         :features
+             :test-paths ["test/features"]}]
+             :kaocha/plugins [:kaocha.plugin.some.required/plugin
+                              ,,,]
+             }
+
+        #include "tests.user.edn"]
+```
+
+This setup would allow developers to maintain their own private
+`tests.user.edn` file and augment the above configuration with for example
+their own set of plugins such as the notifier etc.
+
+The primary advantage to using the `#k/meta-merge` tag over aero's
+`#merge` is that meta-merge will do a deep merge of the tree; meaning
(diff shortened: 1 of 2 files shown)


plexus:
Sorry for losing track of this over the holidays @RickMoynihan. I wanted to merge this but it's marked as draft so GH won't let me. I think this looks good, the only thing I would change is to use just `#meta-merge` instead of `#k/meta-merge`. We don't use an abbreviated `k` prefix anywhere else, and given that we already have all the unprefixed reader literals provided by aero I think it's fine to add this one too. Worst case scenario Aero adds its own `#meta-merge` down the line, in which case I doubt it would have different semantics.

Which makes me think it would make sense to check if the Juxt folks are open to adding `#meta-merge` directly to Aero. wdyt @malcolmsparks @SevereOverfl0w?

SevereOverfl0w:
We've intentionally avoided meta merge. You have to do more mental execution of the code than we are happy with, breaking the obvious principle (Although there's certainly cases where we do that already). I'd say that a meta merge in aero is fairly unlikely at this time.

plexus:
ok, thanks for the input! in that case we'll add it in Kaocha. I would be really useful for this pattern.

``` clojure
;; tests.ci.edn
#meta-merge
[#include "tests.edn"
 {... ci specific config ...}]
```

RickMoynihan:
Thanks for merging!

PR #137, opened 2020-03-10
Support Clojure Expecations For Colourised Output

### LINKED ISSUE #136: Support Clojure Expectations For Colourised Output
Hi,

There is an excellent library called Clojure Expectations that can be found [here](https://github.com/clojure-expectations/clojure-test). It describes itself as `A clojure.test-compatible version of the classic Expectations testing library.` and is actively maintained (unlike the classic library). This library is also 100% compatible with the Clojure Test namespace.

Unfortunately, at the moment, if there are test failures, there is no pretty output from Kaocha like there is with the standard Clojure Test reporting. The output on test failure from Clojure Test is nice colouring of the information to highlight test differences.

Using Expectations, the output is not colourised by Kaocha.

Here is an example from Kaocha using the debug output of a test failure showing a `deftest` output:

```
{:type :fail, :file "foo_test.clj", :line 64, :expected (= true false), :actual (not (= true false)), :message nil, :kaocha/testable #:kaocha.testable{:id :txi.robin.patterns.foo-test/foobar, :type :kaocha.type/var}}
```

and here is one using the Expecations output showing `defexpect` output:

```
{:type :fail, :file "foo_test.clj", :line 61, :expected (=? true false), :actual (not (=? true false)), :message "", :kaocha/testable #:kaocha.testable{:id :txi.robin.patterns.foo-test/foo, :type :kaocha.type/var}}
```

I hope that helps. Let me know if you require further information.

-=david=-

dharrigan:
This commit allows for the `expect` macro in Clojure Expecations to be
supported in Kaocha, starting with the equality operation.

Fixes #136

-=david=-

### DIFF
--- a/src/kaocha/report.clj
+++ b/src/kaocha/report.clj
@@ -237,8 +237,7 @@
                    :actual '~form})
     (t/assert-predicate msg form)))
 
-
-(defmethod print-expr '= [m]
+(defn print-expression [m]
   (let [printer (output/printer)]
     ;; :actual is of the form (not (= ...))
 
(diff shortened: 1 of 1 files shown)


plexus:
Released in `0.0-597`

PR #141, opened 2020-04-23
Switch to using CircleCI's matrix jobs feature

danielcompton:
See https://circleci.com/blog/circleci-matrix-jobs/ and https://circleci.com/docs/2.0/configuration-reference/#matrix for more details.


plexus:
This is nice, we'll have to copy that over to the rest of our projects :)

PR #157, opened 2020-09-12
Fall back to config values in test.check plugin

### LINKED ISSUE #156: kaocha.plugin.alpha.spec-test-check ignores configuration from tests.edn
https://github.com/lambdaisland/kaocha/blob/master/src/kaocha/plugin/alpha/spec_test_check.clj#L56

These lines should be modified not to only read from the CLI-supplied configuration for instrumentation and assertion checking—they should grab these values from `tests.edn` if they're available, because the values specified in `tests.edn` are quietly ignored. To use the configuration from `tests.edn`, one must manually set up a generative fdef check suite. Not a huge deal, but it's something I didn't notice until I was debugging some other test configuration and I saw that these configuration values were unexpectedly set to nil.

I can open a PR that does this when I have some time. Thanks for the great piece of software!

robhanlon22:
* Prefer CLI options over config values
* Enable test.check plugin in tests.edn
* Disable some generative fdef checks for functions without adequate generators

Resolves #156.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -26,6 +26,9 @@ jobs:
             - kaocha/execute:
                 args: "integration --reporter documentation --plugin cloverage --codecov"
                 clojure_version: << parameters.clojure_version >>
+            - kaocha/execute:
+                args: "generative-fdef-checks --reporter documentation"
+                clojure_version: << parameters.clojure_version >>
       - kaocha/upload_codecov:
           flags: integration
           file: target/coverage/integration*/codecov.json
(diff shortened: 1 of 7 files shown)


plexus:
This looks fantastic @robhanlon22! I wish all contributions would be this well rounded. Really appreciate that you added tests, great job fixing the docs, and good idea to enable fdef tests for kaocha itself.

I'll add it this to the CHANGELOG and put out a new release. Feel free to update the CHANGELOG yourself in the future, just add a new entry at the top under `Unreleased`.

The functions marked with `^:no-gen`, is that because they error otherwise?

robhanlon22:
@plexus thank for the high complements! Yes, they are marked with ^:no-gen because they have fn specs that need custom generators.

Issue #159, opened 2020-09-29
test.check.generators FileNotFoundException after upgrading to 1.0.700


vincentjames501:
After updating to `[lambdaisland/kaocha "1.0.700"]`, I started receiving the following:

```bash
$ bin/kaocha                                                                                                                                                                                                                                           
Execution error (FileNotFoundException) at kaocha.specs/fn (specs.clj:82).
Could not locate clojure/test/check/generators__init.class, clojure/test/check/generators.clj or clojure/test/check/generators.cljc on classpath.
```

It seems to have been caused by these additions:
https://github.com/lambdaisland/kaocha/commit/b0c6b1fe795a2b8e91ded58756689dc372fe31ee#diff-2ed781d06f75f1466a634ea0794d8dbe
https://github.com/lambdaisland/kaocha/commit/552b977065786ce90c7cf9e1611943ac4eb88756

honuonhval:
I can also confirm this happens to me.  Fixed it by downgrading to 1.0.690

plexus:
Easiest workaround for now is to add a dependency on test.check

```
[org.clojure/test.check "1.1.0"]
```

honuonhval:
Adding dependency works for me (with 1.0.700).  Thank you

clartaq:
1.0.700 broke my testing too. Adding the `test.check` dependency got it working again.

plexus:
sorry folks, this dependency should be optional, but apparently the code to use it conditionally isn't working as expected, and I don't have a lot of time right now. If someone wants to look at it and make a PR and can put out a new release ASAP.

alysbrooks:
This is also causing an issue with the doc generation. I think it's because cljdocs tries to import the project namespacing in order to read the docstrings.

alysbrooks:
Unfortunately, this isn't completely fixed. 😬 At least one of our specs still requires test.check, and if you use the runner, you often see the problem right away because the problematic spec is used to check the config.

plexus:
This should be fixed now in `lambdaisland/kaocha "1.0.732"`, thanks @alysbrooks for the fix. Could people in this thread please confirm that the latest release fixes this issue for them? Thank!

jr0cket:
@plexus kaocha 1.0.732 is working for my projects, thank you.

alysbrooks:
Considering this closed because we've gotten some confirmation, plus no reports to the contrary.

PR #163, opened 2020-10-20
STC plugin and runner improvements

### LINKED ISSUE #160: Should ::stc/check-asserts? and ::stc/instrument? be specified at the test level?
Right now, `::stc/opts` comes in as part of a test configuration:

https://github.com/lambdaisland/kaocha/blob/29ad4279b45310df5294fb9ee686aee5b71fee41/src/kaocha/plugin/alpha/spec_test_check.clj#L31

But the other options come in as part of the global configuration:

https://github.com/lambdaisland/kaocha/blob/29ad4279b45310df5294fb9ee686aee5b71fee41/src/kaocha/plugin/alpha/spec_test_check.clj#L56
https://github.com/lambdaisland/kaocha/blob/29ad4279b45310df5294fb9ee686aee5b71fee41/src/kaocha/plugin/alpha/spec_test_check.clj#L59

It would be more consistent to choose one or the other. Happy to do the work once a decision is made.

robhanlon22:
- The STC plugin now respects config precedence. CLI options always override test suite and global configuration. Test suite configuration takes precedence over global configuration.
- The STC plugin now reads the suite IDs from config to determine if it needs to skip any config overriding when test selectors are provided, rather than just checking against the static :generative-fdef-checks selector.
- The STC runner now adds test suite configuration to the generated fdef testables. Previously, no options specified within the test suite were respected by the fdef runner.

Resolves #160.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,8 +1,8 @@
 version: 2.1
 
 orbs:
-  kaocha: lambdaisland/kaocha@dev:first
-  clojure: lambdaisland/clojure@dev:first
+  kaocha: lambdaisland/kaocha@0.0.1
+  clojure: lambdaisland/clojure@0.0.1
 
 jobs:
   test:
--- a/src/kaocha/plugin/alpha/spec_test_check.clj
+++ b/src/kaocha/plugin/alpha/spec_test_check.clj
@@ -1,13 +1,42 @@
 (ns kaocha.plugin.alpha.spec-test-check
-  (:require [clojure.spec.alpha]
+  (:require [clojure.spec.alpha :as s]
             [clojure.spec.test.alpha]
             [kaocha.plugin :refer [defplugin]]
             [kaocha.specs]
+            [kaocha.type.spec.test.check]
             [meta-merge.core :refer [meta-merge]]))
 
 ;; This namespace does not actually exist, but is created by
 ;; requiring clojure.spec.test.alpha
 (alias 'stc 'clojure.spec.test.check)
+
+(s/def ::config-path
+  (s/coll-of keyword? :min-count 1))
+
+(s/def ::config-mapping
+  (s/map-of ::config-path ::config-path))
+
+(s/def ::stc-config
+  (s/keys :opt [::stc/instrument?
+                ::stc/check-asserts?
+                ::stc/opts]))
+
+(s/def ::cli ::stc-config)
+(s/def ::global ::stc-config)
+
+(s/def ::stc-configs
+  (s/keys :opt [::cli ::global]))
+
+(s/def ::test
+  (s/or :stc-test   :kaocha.type/spec.test.check
+        :other-test :kaocha/testable))
+
+(s/def ::tests
+  (s/coll-of ::test))
+
+(s/fdef is-stc?
+  :args (s/cat :test ::test)
+  :ret  boolean?)
 
 (defn ^:private is-stc?
   [test]
@@ -26,12 +55,20 @@
    [::stc/opts :num-tests]
    [::stc/opts :max-size]])
 
-(def ^:private cli-paths
+(def ^:private cli-config-paths
   (mapv (partial vector :kaocha/cli-options)
         [:stc-instrumentation
          :stc-asserts
          :stc-num-tests
          :stc-max-size]))
+
+(s/fdef make-config-mapping
+  :args (s/cat :instrument?-path    ::config-path
+               :check-asserts?-path ::config-path
+               :opts-num-tests-path ::config-path
+               :opts-max-size-path  ::config-path)
+  :ret  (s/map-of ::config-path ::config-path
+                  :count (count config-paths)))
 
 (defn ^:private make-config-mapping
   [instrument?-path check-asserts?-path opts-num-tests-path opts-max-size-path]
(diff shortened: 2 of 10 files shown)


src/kaocha/plugin/alpha/spec_test_check.clj
```diff
@@ -46,24 +86,10 @@
           [nil "--stc-max-size SIZE"        "spec.test.check: Maximum length of generated collections"
            :parse-fn #(Integer/parseInt %)]))
   (config [{:kaocha/keys [tests cli-args] :as config}]
-    (if (and (seq cli-args)
-             (not (some #{:generative-fdef-checks} cli-args)))
+    (if (and (seq cli-args) (not (some #{:generative-fdef-checks} cli-args)))
```
robhanlon22, line 89: Something I noticed here: this should probably use the `:kaocha.testable/id`s specified by all specified `generative-fdef-check` suites instead of this hard-coded ID.

plexus:
Is this good to go @robhanlon22 ? Would you mind adding an entry to the CHANGELOG? Thanks!

robhanlon22:
Oh, sure! Also, I thought about this some more, and I noticed that the global config would override any specific test config. I'll update this to prefer the test suite config.

robhanlon22:
@plexus code complete and CHANGELOG updated!

robhanlon22:
@plexus I added another commit that dogfoods the plugin by writing specs for it. It turns out that there were some spec definitions missing or misplaced; I've defined/moved them appropriately.

robhanlon22:
@plexus also, it looks like there's a CircleCI error that I can't fix:

```
#!/bin/sh -eo pipefail
# The dev version of lambdaisland/kaocha@dev:first has expired. Dev versions of orbs are only valid for 90 days after publishing.
# The dev version of lambdaisland/clojure@dev:first has expired. Dev versions of orbs are only valid for 90 days after publishing.
# 
# -------
# Warning: This configuration was auto-generated to show you the message above.
# Don't rerun this job. Rerunning will have no effect.
false

Exited with code exit status 1
CircleCI received exit code 1
```

src/kaocha/plugin/alpha/spec_test_check.clj
```diff
@@ -1,40 +1,177 @@
...
+(s/fdef config-hook
+  :args (s/cat :config :kaocha/config)
+  :ret  :kaocha/config)
+
+(defn ^:private config-hook
+  [config]
+  (let [config-with-suite (update config :kaocha/tests ensure-stc-suite)]
+    (if (stc-suites-not-selected? config-with-suite)
+      config
+      (override-stc-config config-with-suite))))
```
robhanlon22, line 174: `fdef` can be a little verbose. Orchestra's [defn-spec](https://github.com/jeaye/orchestra#defn-spec) is some nice sugar around it—thoughts?

src/kaocha/specs.clj
```diff
@@ -95,18 +120,23 @@
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;; clojure.spec.test
 
+(try
+  (require 'clojure.spec.test.alpha)
+  (alias 'stc 'clojure.spec.test.check)
+  (catch FileNotFoundException _))
```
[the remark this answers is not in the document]
plexus: Does this fix https://github.com/lambdaisland/kaocha/issues/159 ?

src/kaocha/type/spec/test/ns.clj
```diff
@@ -27,14 +27,14 @@
 
 (stest/checkable-syms)
 
-(type.fdef/load-testables '[kaocha.result/sum])
+(type.fdef/load-testables '[kaocha.result/sum] {})
```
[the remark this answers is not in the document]
plexus: Probably not... feel free to take it out. I guess we'll find out if we miss it :)

src/kaocha/plugin/alpha/spec_test_check.clj
plexus: Honestly I don't use spec and fdef that much so no strong opinions. Since we already have orchestra as a dependency I think it's fine to use it.

.circleci/config.yml
```diff
@@ -1,8 +1,8 @@
 version: 2.1
 
 orbs:
-  kaocha: lambdaisland/kaocha@dev:first
-  clojure: lambdaisland/clojure@dev:first
+  kaocha: lambdaisland/kaocha@0.0.1
+  clojure: lambdaisland/clojure@0.0.1
```
plexus, line 5: :+1:

src/kaocha/specs.clj
```diff
@@ -1,27 +1,40 @@
...
+  (s/keys :opt [:kaocha/reporter
+                :kaocha/color?
+                :kaocha/fail-fast?
+                :kaocha/watch?
+                :kaocha/plugins]))
+
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;; config
 
-(s/def :kaocha/config (s/keys :req ~(conj global-opts :kaocha/tests)))
```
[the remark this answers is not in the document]
plexus: That probably would make sense. When we read `tests.edn` there's normally a `#kaocha/v1` which does the processing/normalizing, so you should have a complete config map after that. But it would be useful to give people feedback on their config before processing, and their are cases where we get unprocessed or partial config maps, like when calling `kaocha.repl`.

tests.edn
```diff
@@ -21,5 +22,13 @@
...
  :clojure.spec.test.check/instrument? true
  :clojure.spec.test.check/check-asserts? true
+ :clojure.spec.test.check/opts
+ #profile {;; Halving the default opts to prevent CircleCI from killing the test
+           ;; process for lack of output.
+           :ci {:num-tests 500 :max-size 100}
+           ;; Reducing num-tests by two orders of magnitude in local
+           ;; development greatly speeds up iteration time at the expense of
+           ;; possible false negatives.
+           :default {:num-tests 10}}
```
plexus, line 32: Nice use of #profile !

plexus:
Thanks @robhanlon22 , I appreciate the effort, and seems like this will contain some solid improvements. Let me know if you're done with this or not and I'll push the button and put out a release.

robhanlon22:
We’re there! Sorry for the churn, @plexus, I was having a good time :)

plexus:
the-big-lebowski-dancing-o

Thanks!

plexus:

PR #165, opened 2020-11-04
Specs for raw configuration values

robhanlon22:
This is a continuation of #163. This add more spec definitions, specifically around raw, unprocessed configuration, where much of the data can be optionally specified and specified without qualification. With this, a few more functions can have fdefs written for them, and it makes a clearer delineation between input configuration and internal configuration.

### DIFF
--- a/src/kaocha/specs.clj
+++ b/src/kaocha/specs.clj
@@ -39,7 +39,7 @@
 ;; config
 
 
-(s/def :kaocha/bindings (s/map-of qualified-symbol? (s/coll-of keyword?)))
+(s/def :kaocha/bindings (s/map-of qualified-symbol? any?))
 
 (s/def :kaocha.testable/meta (s/nilable map?))
 
--- a/src/kaocha/specs.clj
+++ b/src/kaocha/specs.clj
@@ -53,9 +53,9 @@
 
 (s/def :kaocha/ns-patterns (s/coll-of string?))
 
-(s/def :kaocha.filter/skip (s/coll-of symbol?))
+(s/def :kaocha.filter/skip (s/coll-of keyword?))
 
-(s/def :kaocha.filter/focus (s/coll-of symbol?))
+(s/def :kaocha.filter/focus (s/coll-of keyword?))
 
 (s/def :kaocha.filter/skip-meta (s/coll-of keyword?))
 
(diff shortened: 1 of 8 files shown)


src/kaocha/specs.clj
```diff
@@ -44,10 +77,89 @@
...
+
+(s/def :kaocha.plugin.randomize/randomize?
+  any?)
+
+(s/def :kaocha.plugin.capture-output/capture-output?
+  any?)
+
+(s/def :kaocha/tests (s/coll-of :kaocha/testable))
+
+(s/def :kaocha/testable (s/keys :opt [:kaocha.testable/type
```
robhanlon22, line 89: I'm wondering if this should be decomposed further into a "test" from config and a "testable" that is the result of load-testables.

plexus: Don't we already have that? `:kaocha/testable` vs `:kaocha.test-plan/testable`?

plexus:
Looks good at first glance, sorry for the delay. Could you see what's up with the failing tests?

src/kaocha/specs.clj
robhanlon22: oh, yes! my bad.

plexus:
@robhanlon22 I took the liberty to push a few commits. The tests were getting stuck, in particular the watch-test, and it was a bit hard to see why. The downside of a test runner testing itself, we're redirecting output all over the place and so sometimes things get hidden that you would rather want to have in your face.

Anyway with some poking around I managed to refine the specs a bit

- bindings can really be anything, we just set these as dynamic bindings, we don't care about the value. Only the key needs to be a valid var name (qualified-symbol?)
- focus and skip are vectors of test-ids, i.e. keywords. Setting it to symbol? broke usage of `--focus`, but setting it to keyword? broke some of our integration tests. It seems so far we have allowed both in tests.edn, I've changed kaocha.config to normalize them to keywords.

Let's see how this fares on CI now. Could you add a changelog entry?

Note that you can run things locally with the CI profile with a simple `CI=true bin/kaocha`.

robhanlon22:
Thanks @plexus! I'll try to get back around to this when things cool down a bit at work :)

plexus:
```
bin/kaocha  --focus 'command-line.print-config/using---print-config'
```

This test is failing now, it needs to be updated to show the new normalized config map.

alysbrooks:
@robhanlon22 I'm looking at older issues/PRs and cleaning them up. Are you interested in picking this up again? If not I might try to wrap this up/see if @plexus is interested in wrapping it up.

Issue #167, opened 2020-11-13
Add explanation of defaults from README to Docs


alysbrooks:
I tried Kaocha on my own project and after installation, I was unclear on whether configuration was necessary, recommended, or optional. We do [a great job clarifying this in the README](https://github.com/lambdaisland/kaocha#all-tools), but I missed the Quickstart section. I think we should explain it in the Configuration section of the docs as well.

plexus:
Thanks for fixing this!

PR #168, opened 2020-11-13
Copy useful information about defaults from README to the Configuration

alysbrooks:
Addressing #167, I took some information from the README and copied it to the Configuration section.

Obviously, we don't want too much duplication in the docs, but I think this is important enough for newcomers to Kaocha to have in both places.


plexus:
Thanks, this is great!

plexus:
peanuts

Issue #36, opened 2019-01-18
Warn if plugin name in config is not a keyword


plexus:
Plugin names are assumed to be keywords (it's an identifier for the plugin). However it's easy to mix this up and put in a symbol. We could allow this and coerce, but I'd rather enforce systematic use of keywords. That also has the benefit that symbols are still available for some later extension.

This isn't currently caught by spec I think because we let plugins run their config hooks first before validating the config, but maybe we should really validate twice, once before and once after the config hook.

riyasingh1004:
Hi , would like to work on this issue. Please help me get started with it.

riyasingh1004:
I will like to work on this issue. Please help me get started.

plexus:
Hi @riyasingh1004 :wave: great to have you here!

I had a better look at what needs to happen here. There are two things that are missing.

1. We need to add a spec for `:kaocha/plugins`
2. We need to validate the configuration

Step one is actually very straightforward, in `src/kaocha/specs.clj`, we need something like this. You can add it at the top somewhere, for instance around line 14.

``` clojure
(s/def :kaocha/plugins (s/coll-of keyword?))
```

What this does is tell clojure.spec that if a map contains a `:kaocha/plugins` key, then the value corresponding with that key has to be a collection of keywords. If you are not familiar yet with clojure spec then I recommend watching [this introduction video](https://lambdaisland.com/episodes/clojure-spec) and trying it out. Ping me if you have questions.

So that is step one. We already have a spec saying that a kaocha configuration needs to contain `:kaocha/plugins` so that's good.

The second step is to actually validate the configuration so the user gets feedback. You can use [expound](https://github.com/bhb/expound) to create really friendly and helpful messages, this is what it would look like:

```
 % bin/kaocha                                                                                                                                                                                                                                                                                   ✘ 252 
Invalid configuration in tests.edn
-- Spec failed --------------------

  {:kaocha/reporter ...,
   :kaocha/color? ...,
   :kaocha/fail-fast? ...,
   :kaocha/plugins [... ... ... foo.bar ...],
                                ^^^^^^^
   :kaocha/tests ...,
   :kaocha.hooks/pre-run ...}

should satisfy

  keyword?

-- Relevant specs -------

:kaocha/plugins:
  (clojure.spec.alpha/coll-of clojure.core/keyword?)
:kaocha/config:
  (clojure.spec.alpha/keys :req ~(conj global-opts :kaocha/tests))

-------------------------
Detected 1 error
[...]

PR #169, opened 2020-11-14
Add spec validation for the list of plugins.

alysbrooks:
Addresses #36 using @plexus's suggested solution of adding a spec validation check early on.

I added a simple exception handler to just print the exception name and spec failed message, so it looks like the following:

```
java.lang.AssertionError: -- Spec failed --------------------

  {:kaocha/bindings ...,
   :kaocha/tests ...,
   :kaocha/fail-fast? ...,
   :kaocha/color? ...,
   :clojure.spec.test.check/instrument? ...,
   :kaocha/plugins
   [... ... ... kaocha.plugin.alpha/info ... ... ... ... ...],
                ^^^^^^^^^^^^^^^^^^^^^^^^
   :kaocha/reporter ...,
   :clojure.spec.test.check/check-asserts? ...,
   :kaocha.hooks/pre-load ...}

should satisfy

  keyword?

-- Relevant specs -------

:kaocha/plugins:
  (clojure.spec.alpha/coll-of clojure.core/keyword?)
:kaocha/config:
  (clojure.spec.alpha/keys :req ~(conj global-opts :kaocha/tests))

-------------------------
Detected 1 error
```

It looks like Kaocha code base uses `(throw+ {:kaocha/early-exit 0})` in this circumstance, although I'm not sure if the error code is the one we want.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,8 +1,8 @@
 version: 2.1
 
 orbs:
-  kaocha: lambdaisland/kaocha@dev:first
-  clojure: lambdaisland/clojure@dev:first
+  kaocha: lambdaisland/kaocha@0.0.1
+  clojure: lambdaisland/clojure@0.0.1
 
 jobs:
   test:
--- a/src/kaocha/runner.clj
+++ b/src/kaocha/runner.clj
@@ -154,7 +154,11 @@
                                                               (config/load-config (if profile
                                                                                     {:profile profile}
                                                                                     {})))
-          _check (specs/assert-spec :kaocha/config config)
+          _check                                          (try 
+                                                            (specs/assert-spec :kaocha/config config)
+                                                            (catch AssertionError e 
+                                                              (println (.toString e))
+                                                              (throw+ {:kaocha/early-exit 0})))
           plugin-chain                                    (plugin/load-all (concat (:kaocha/plugins config) plugin))
           cli-options                                     (plugin/run-hook* plugin-chain :kaocha.hooks/cli-options cli-options)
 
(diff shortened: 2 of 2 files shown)


plexus:
> `(throw+ {:kaocha/early-exit 0})`

This is the process exit code, since this is a non-successful exit it should be non-zero. So far it seems we're using `254` and `253` already, so maybe `252`. It would also be a good idea to document these, so you can check them for instance in a shell script that calls kaocha.

alysbrooks:
I updated it to use a nonzero error code and tweaked the error message a little:

```
ERROR: Invalid configuration file:
-- Spec failed --------------------

  {:kaocha/bindings ...,
   :kaocha/tests ...,
   :kaocha/fail-fast? ...,
   :kaocha/color? ...,
   :clojure.spec.test.check/instrument? ...,
   :kaocha/plugins
   [... ... ... kaocha.plugin.alpha/info ... ... ... ... ...],
                ^^^^^^^^^^^^^^^^^^^^^^^^
   :kaocha/reporter ...,
   :clojure.spec.test.check/check-asserts? ...,
   :kaocha.hooks/pre-load ...}

should satisfy

  keyword?

-- Relevant specs -------

:kaocha/plugins:
  (clojure.spec.alpha/coll-of clojure.core/keyword?)
:kaocha/config:
  (clojure.spec.alpha/keys :req ~(conj global-opts :kaocha/tests))

-------------------------
Detected 1 error
```

(Error will actually appear in red in terminals configured to display color.)

plexus:
Great! In the future please add an entry to the `CHANGELOG.md`. We should probably add a PR template that mentions that, since I'm constantly reminding people :)

plexus:

Issue #166, opened 2020-11-13
Clarify how to use REPL from Leningen


alysbrooks:
If you follow the installation setup, Kaocha isn't available from the Leiningen REPL because you're creating a separate profile. Instead of `lein repl`, you need to do `lein with-profile +kaocha repl`.

Alternatively, we could implement  #118. I'm listing this as a separate issue in case we want to keep the Kaocha profile around as an alternative option.

plexus:
Let's go with #118 , people can still put it into its own profile if they like, but that way the default setup is a little more foolproof.

alysbrooks:
Fixed by #170 ✨

Issue #118, opened 2019-10-10
Default to the dev profile?


AndreaCrotti:
When I started using Kaocha I followed the suggestion to add all the dependencies into the "kaocha" profile.e

This is a good suggestion, but I wonder if it should be the default.
I stopped using the kaocha profile, just because it makes it a bit less convenient for everyone to have kaocha always available (in the Repl for example)
It's still possible of course with the right parameters passed to Leiningen for example, but imho it doesn't really harm to have it in the dev profile rather than the Kaocha profile.

Any thoughts?

alysbrooks:
I agree that the dev profile is a better default. I think we should retain the description of making it a separate profile for people who either:

- Rarely or never use Kaocha in the REPL
- Only use Kaocha for a specific purpose, e.g., if they usually use the default test runner but use Kaocha for test profiling.

plexus:
I think that makes sense, with Clojure CLI I tend to stick kaocha in the `:test` alias, which I make sure to include when starting a REPL. People should have Kaocha available when working on the REPL by default, so I think for Leiningen recommending the `:dev` profile is a good idea.

alysbrooks:
Since we're agreed, I can update the docs.

alysbrooks:
Thanks! I updated our docs to recommend the `:dev` profile, so I think this is addressed.

PR #170, opened 2020-11-18
Update docs to suggest using :dev profile for Leiningen

alysbrooks:
Since a test runner is a core part of the dev workflow, it makes more sense to add Kaocha as a dependency to `:dev` than as as separate profile. 

It does sometimes make sense to use a separate profile, so I'm keeping that information but putting it in a separate section that clarifies the tradeoffs.

Addresses #118 and #166.


plexus:
thankyou

PR #171, opened 2020-11-18
Make sure we can still run Kaocha when test.check is not available

### LINKED ISSUE #159: test.check.generators FileNotFoundException after upgrading to 1.0.700

plexus:
Use wrappers for `s/gen` and `s/with-gen`, when we are unable to load
test.check.generators we use simple stubs.

Fixes #159

### DIFF
--- a/src/kaocha/specs.clj
+++ b/src/kaocha/specs.clj
@@ -4,8 +4,13 @@
             [expound.alpha :as expound])
   (:import (java.io FileNotFoundException)))
 
+(defn s-gen [_])
+(defn s-with-gen [spec _] spec)
+
 (try
   (require 'clojure.test.check.generators)
+  (def s-gen @(resolve 'clojure.spec.alpha/gen))
+  (def s-with-gen @(resolve 'clojure.spec.alpha/with-gen))
   (catch FileNotFoundException _))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
@@ -61,9 +66,9 @@
 (s/def :kaocha.testable/desc string?)
 
 (s/def :kaocha.testable/wrap
-  (s/with-gen
-    (s/coll-of fn? :into [])
-    #(s/gen #{[]})))
+  (s-with-gen
+   (s/coll-of fn? :into [])
+   #(s-gen #{[]})))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;; Test plan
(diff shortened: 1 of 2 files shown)


alysbrooks:
Looks good. 

I did run into this error when trying to run kaocha tests using this branch without having test.check on my classpath:

```
--- unit (clojure.test) ---------------------------Execution error (FileNotFoundException) at orchestra.spec.test/spec-checking-fn$conform! (test.cljc:15).
Could not locate clojure/test/check/generators__init.class, clojure/test/check/generators.clj or clojure/test/check/generators.cljc on classpath.
```

I don't know if that's arising from a place in kaocha that depends on test.check and shouldn't or is just one of kaocha's tests that depends on test.check. I think it might be the latter and not a problem? In any case, this PR already fixes many of the places I've noticed #159, so perhaps that shouldn't hold up merging this.

One suggestion: Would it make more sense to put the definitions of `s-gen` and `s-with-gen` inside the `catch`? I think they're functionally identical, but it might be clearer. Also, it reduces the odds of the fallback definitions being separated from the regular ones.

alysbrooks (commented):
Changes look good! I did have a suggestion I made in line.

I did run into this error when trying to run kaocha tests using this branch without having test.check on my classpath:

```
--- unit (clojure.test) ---------------------------Execution error (FileNotFoundException) at orchestra.spec.test/spec-checking-fn$conform! (test.cljc:15).
Could not locate clojure/test/check/generators__init.class, clojure/test/check/generators.clj or clojure/test/check/generators.cljc on classpath.
```

I don't know if that's arising from a place in kaocha that depends on test.check and shouldn't or is just one of kaocha's tests that depends on test.check. I think it might be the latter and not a problem? In any case, this PR already fixes many of the places I've noticed #159, so perhaps that shouldn't hold up merging this.

src/kaocha/specs.clj
```diff
@@ -4,8 +4,13 @@
...
   (:import (java.io FileNotFoundException)))
 
+(defn s-gen [_])
+(defn s-with-gen [spec _] spec)
+
 (try
   (require 'clojure.test.check.generators)
+  (def s-gen @(resolve 'clojure.spec.alpha/gen))
+  (def s-with-gen @(resolve 'clojure.spec.alpha/with-gen))
   (catch FileNotFoundException _))
```
alysbrooks, line 14: Would it make more sense to put the definitions of s-gen and s-with-gen here? I think they're functionally identical, but it might be clearer. Also, it reduces the odds of the fallback definitions being separated from the regular ones.
```suggestion
  (catch FileNotFoundException _
  (defn s-gen [_])
  (defn s-with-gen [spec _] spec)))
```

plexus: I think having them at the top level may be beneficial for people using tools that rely on static analysis (i.e. Cursive). Not sure if it's smart enough to infer that `s-gen` and `s-with-gen` exist otherwise.

PR #172, opened 2020-11-18
docs: change `use` to `require` in REPL example

oxalorg:
Problem: Docs for using koacho with REPL is using `use`

Solution: avoid `use` and be more explicit with `require`

Optionally example could have used `(requrie '[kaocha.repl :as k :refer [run]])` but `run` is a common name which may exist in users namespace.


alysbrooks:
I agree, referring `run` doesn't make a lot of sense.

Looks good to merge!

PR #174, opened 2020-11-25
Ensure runner works without test.check.

alysbrooks:
Users are still running into issue #159. It happens soon after starting `bin/kaocha` because the problematic spec is used to check the config.

I narrowed it down, and the problem is in the`kaocha/report` spec. `clojure.spec.alpha.fspec` actually requires test.check, which is [a known limitation](https://clojure.atlassian.net/browse/CLJ-1936#icft=CLJ-1936). ~~To fix this, I used the same technique as in #171 .~~ (see comment below)


alysbrooks:
Using a simple `def` doesn't work with fspec, which is a macro. I tried a few suggestions I found online for aliasing a macro, but it turns out simply using `defmacro` is the most reliable.

plexus:

Issue #37, opened 2019-01-18
Fail fast if clojure version < 1.9


plexus:
We need 1.9 because of spec, we should find some way to communicate that immediately to the user, rather than erroring out on some failed require statement.

Same for Kaocha-cljs, which relies on ClojureScript 1.10 (and thus Clojure 1.10).

danielcompton:
Does the spec backport library work, or is the spec use more integral here?

plexus:
I have no idea, I never tried it. Does it also implement the extra predicates like `qualified-keyword?`? Is it compatible with Orchestra and Expound?

If someone wants to explore this that would of course be great, but I am not particularly invested in pursuing pre 1.9 compatibility. Say we manage to be 1.8-compatible and start getting more users, then we'll likely be tied to 1.8 for years to come, as I'd like to avoid dropping existing users. So I'd rather draw the line at 1.9, and try to keep 1.9 compatibility for the foreseeable future. (at least until there are a few more major Clojure versions out, and there's a compelling reason to get onto newer features).

danielcompton:
Limiting support to 1.9 seems like a pretty reasonable default at the moment anyway, so it's probably not worth looking at too closely.

alysbrooks:
I briefly tried the spec-backport library, and it wasn't enough to avoid errors—we'd have to backport other new functionality from 1.9, like `boolean?`.

plexus:
I wouldn't put any effort into pre 1.9 compatibility at this point. The last Clojure Survey shows that close to 90% of the community is on 1.9 or 1.10, and that number will only increase.

https://clojure.org/news/2020/02/20/state-of-clojure-2020

alysbrooks:
Should be addressed by #177

PR #177, opened 2020-11-27
Throw error when on a version of Clojure prior to 1.9.

alysbrooks:
Addresses #37. Checking for the Clojure version is a bit tricky because the initial errors come during the import process, so they can't be handled during the normal program flow. 

The only alternative I came up with would be to enclose problematic imports in `try`s. Since we don't support earlier versions, littering the codebase with checks for it seems like a bad tradeoff.

### DIFF
--- a/src/kaocha/api.clj
+++ b/src/kaocha/api.clj
@@ -1,4 +1,4 @@
-(require 'kaocha.version_check)
+(require 'kaocha.version-check)
 (ns kaocha.api
   "Programmable test runner interface."
   (:require [clojure.test :as t]
--- a/src/kaocha/api.clj
+++ b/src/kaocha/api.clj
@@ -1,3 +1,4 @@
+(require 'kaocha.version_check)
 (ns kaocha.api
   "Programmable test runner interface."
   (:require [clojure.test :as t]
--- a/src/kaocha/core_ext.clj
+++ b/src/kaocha/core_ext.clj
@@ -5,6 +5,7 @@
   (:require  [kaocha.output :as output]
             [slingshot.slingshot :refer [try+ throw+]]))
 
+;This ends up being the best place to put the version check to avoid duplication.
 (when-not (and (>= (:major *clojure-version*) 1) (>= (:minor *clojure-version*) 9))
                        (output/error "Kaocha requires Clojure 1.9 or later.")
                        (throw+ {:kaocha/early-exit 251}))
--- a/src/kaocha/core_ext.clj
+++ b/src/kaocha/core_ext.clj
@@ -6,7 +6,8 @@
             [slingshot.slingshot :refer [try+ throw+]]))
 
 ;This ends up being the best place to put the version check to avoid duplication.
-(when-not (and (>= (:major *clojure-version*) 1) (>= (:minor *clojure-version*) 9))
+(when-not (or (and (= (:major *clojure-version*) 1) (>= (:minor *clojure-version*) 9))
+              (>= (:major *clojure-version*) 2))
                        (output/error "Kaocha requires Clojure 1.9 or later.")
                        (throw+ {:kaocha/early-exit 251}))
 
--- a/src/kaocha/core_ext.clj
+++ b/src/kaocha/core_ext.clj
@@ -3,7 +3,6 @@
   (:refer-clojure :exclude [symbol])
   (:import [java.util.regex Pattern]))
 
-
 (defn regex? [x]
   (instance? Pattern x))
 
--- a/src/kaocha/core_ext.clj
+++ b/src/kaocha/core_ext.clj
@@ -1,15 +1,8 @@
 (ns kaocha.core-ext
   "Core language extensions"
   (:refer-clojure :exclude [symbol])
-  (:import [java.util.regex Pattern])
-  (:require  [kaocha.output :as output]
-            [slingshot.slingshot :refer [try+ throw+]]))
+  (:import [java.util.regex Pattern]))
 
-;This ends up being the best place to put the version check to avoid duplication.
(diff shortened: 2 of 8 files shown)


src/kaocha/core_ext.clj
```diff
@@ -1,7 +1,14 @@
 (ns kaocha.core-ext
   "Core language extensions"
   (:refer-clojure :exclude [symbol])
-  (:import [java.util.regex Pattern]))
+  (:import [java.util.regex Pattern])
+  (:require  [kaocha.output :as output]
+            [slingshot.slingshot :refer [try+ throw+]]))
+
+;This ends up being the best place to put the version check to avoid duplication.
+(when-not (and (>= (:major *clojure-version*) 1) (>= (:minor *clojure-version*) 9))
```
plexus, line 9: If Clojure ever gets a 2.0 this is going to break :)

src/kaocha/runner.clj
```diff
@@ -1,14 +1,14 @@
 (ns kaocha.runner
   "Main entry point for command line use."
   (:gen-class)
-  (:require [clojure.java.io :as io]
+  (:require [kaocha.api :as api]
+            [clojure.java.io :as io]
```
plexus, line 5: Is this necessary because it's influencing the load order? If so then we should put a big fat comment around it, since I will often let Emacs automatically sort my ns declerations, so this will inadvertently break probably as soon as I touch this namespace.

alysbrooks: I added a comment. This still seems a bit too fragile to me. What if I created a separate file just for the version check and `require`d  it before the namespace declaration?

src/kaocha/core_ext.clj
alysbrooks: Ahh, right of course. Fixed! :)

src/kaocha/runner.clj
plexus: I think that's a good idea!

alysbrooks:
Made the change and added tests. Having a separate function and tests for it might have been overkill, but I already had examples in my REPL, so it seemed pragmatic to convert them into actual tests. Plus, having a check version function allows us or third-party plugin authors to ship plugins with a higher version requirement than Kaocha as a whole.

src/kaocha/api.clj
```diff
@@ -1,3 +1,4 @@
+(require 'kaocha.version_check)
```
plexus, line 1: can we use `version-check` instead of `version_check`? I think that would be more idiomatic and future-proof.

This "require before ns" trick is interesting. I'm sure some static analysis tools are not going to like it, but I think it's fine. We'll see if anyone complains.

plexus:
Looks good to me! let me know if you still want to make changes, if not then this can go in.

Oh could you try to add CHANGELOG entries in the future? not a big effort to add them before a release but it's even handier if they get added inside the PR.

alysbrooks:
I can add a CHANGELOG, but once I do that and change the name, it's ready.

src/kaocha/version_check.clj
```diff
@@ -0,0 +1,14 @@
...
+(defn check-version-minimum 
+  "Checks that Clojure has at least a minimum version"
+  [major minor]
+  (when-not (or (and (= (:major *clojure-version*) major) (>= (:minor *clojure-version*) minor))
+                (>= (:major *clojure-version*) (inc major) ))
+    (let [msg (format "Kaocha requires Clojure %d.%d or later." major minor)]
+      (output/error msg)
+      (throw+ {:kaocha/early-exit 251} msg))))
+
+(check-version-minimum 1 9)
```
plexus, line 14: Maybe wrap this in a `(comment)`

test/unit/kaocha/version_check_test.clj
```diff
@@ -0,0 +1,22 @@
+
+(ns kaocha.version-check-test
+  (:require [clojure.test :refer :all]
+            [kaocha.test-helper]
+            [kaocha.version-check :as v]))
+
+
```
plexus, line 7: Not a blocker, but please be mindful of use of whitespace

https://nextjournal.com/lambdaisland/clojure-style-guide#whitespace

src/kaocha/version_check.clj
alysbrooks: This is actually needed for the check to happen upon importing `kaocha.version-check`. I could change each file to specifically call this function, but since 1.9 is needed for Kaocha as a whole, I'm not sure it's worth it.

test/unit/kaocha/version_check_test.clj
alysbrooks: Thanks for the reminder!

src/kaocha/version_check.clj
plexus: Of course 🤦 never mind me!

test/unit/kaocha/version_check_test.clj
alysbrooks: I fixed this. Let me know if you want me to squash some commits, since a commit just fixing whitespace is pretty trivial.

plexus:
I fixed some more whitespace issues, I see you have a tendency to add a newline at the beginning of the file. Some of the indentation was also a little wonky... https://github.com/lambdaisland/kaocha/pull/177/commits/2f2d05d79365332c0e12f29ca68369f284d1d1d3

Looks good though! There was a merge conflict on the CHANGELOG file, which I resolved through github's UI, which apparently merges `main` back into this branch :facepalm: as mentioned on Discord I'm really not a fan of that, so I guess I'll do my conflict resolution locally again in the future.

alysbrooks:
I think the newlines are coming from places where I added code and then deleted it, but didn't delete the newline I originally added to separate the new code from the existing code. I'll be more careful with that in the future, and maybe look at automatically running some sort of script to warn on extra whitespace when committing.

Issue #179, opened 2020-11-30
Error when passing --focus with multiple test ids


AndreaCrotti:
With multiple test ids, currently passing `--focus` actually goes through all of them.
That's unlikely to be what you actually want and is just confusing, kaocha could just throw an error explaining that you need to pass the test-id as well.

plexus:
It seems like we are talking past each other here, from what you said on Slack I understood that you wanted `--focus` without an argument to be an error, but that is already the case:

```
Missing required argument for "--focus SYM"
```

After reading this issue a few times and looking at your PR I see now that you're addressing something else, the fact that kaocha needs to scan the full test plan to find what to focus on. This is sometimes undesirable, but there's no easy solution for this.

It seems you want to make it impossible for people to use `--focus` without also specifying the test-suite, that's not desirable. You may want to `bin/kaocha --focus test-from-one-suite --focus test-from-other-suite`. You also shouldn't have to specify the test suite, that's never been a requirement and we're not going to change that.

AndreaCrotti:
Yeah ok understood, I can probably close this then.

PR #180, opened 2020-11-30
throw an assertion if --focus is set and there are multiple test ids set

AndreaCrotti:
Closes https://github.com/lambdaisland/kaocha/issues/179

That's a first attempt at least, I guess I need to add a test and/or fixing existing tests.

### DIFF
--- a/src/kaocha/plugin/filter.clj
+++ b/src/kaocha/plugin/filter.clj
@@ -102,6 +102,14 @@
 
   (config [config]
     (let [{:keys [skip focus skip-meta focus-meta]} (:kaocha/cli-options config)]
+      (when (and (or (seq focus)
+                     (seq focus-meta)
+                     (seq skip)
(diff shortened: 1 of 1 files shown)


alysbrooks:
Thanks for the PR! The CircleCI issue isn't because of a problem with your code, it's because you're working off an old version of the code that has an outdated configuration. But additional tests for the multiple test-ids feature would be great.

AndreaCrotti:
Should it fail for all the options like focus/focus-meta and skip/skip-meta?
I think in all these cases you probably need to specify a test id right?

AndreaCrotti:
Another thing I was thinking is that to be smarter, this maybe should fail only if your test would actually run multiple times.
For example if there are clojure and clojurescript tests and they have entirely different test paths anyway, they don't overlap so maybe in that case you should not need to specify the test id?

PR #181, opened 2020-12-01
Better plugin errors

alysbrooks:
Addresses #35. 

What's left to do:

- [x] Extra tests to cover the various error cases
- [x] Refactor `try-load-third-party-lib` or `register`?

### DIFF
--- a/src/kaocha/plugin.clj
+++ b/src/kaocha/plugin.clj
@@ -31,31 +31,31 @@
   (fn [name plugins] name))
 
 (defmethod -register :default [name plugins]
-  (throw+ {:kaocha/early-exit 254} nil (str "Couldn't load plugin " name)))
+  (throw+ {:kaocha.plugin/not-loaded name} nil (str "Couldn't load plugin " name)))
 
 (defn register [plugin-name plugin-stack]
   (let [ns-result (try-load-third-party-lib plugin-name)
         successful-ns (map second (filter first ns-result))
         failed-ns (map second (filter (complement first) ns-result))
         plugin-result (try+ (-register plugin-name plugin-stack)
-                            (catch [:kaocha/early-exit 254] _ false)) ] 
+                            (catch [:kaocha.plugin/not-loaded plugin-name] _ false)) ] 
     (cond
       ;Namespaces succeeded, but plugin itself failed.
       (and (not plugin-result)
            (every? true? (map first ns-result))) (output/error-and-throw {:kaocha/early-exit 254} nil
-                                                  (format "Couldn't load plugin %s, but loaded %s." 
+                                                  (format "Couldn't load plugin %s. The plugin was not defined after loading namespace %s. Is the file missing a defplugin?" 
                                                           plugin-name (first successful-ns))) 
       ;Multiple namespaces failed to load.
       (and (not plugin-result)
            (> (count failed-ns) 1)) (output/error-and-throw 
                                       {:kaocha/early-exit 254} nil
-                                      (format "Couldn't load plugin %s. Failed to load namespaces %s." 
+                                      (format "Couldn't load plugin %s. Failed to load namespaces %s. This could be caused by a misspelling or a missing dependency." 
                                               plugin-name (apply str ( interpose " and " failed-ns))))
       ;A single namespace failed to load. 
       ;This is a separate case mostly so we can get the error message's grammar right.
       (not plugin-result) (output/error-and-throw 
                             {:kaocha/early-exit 254} nil
-                            (format "Couldn't load plugin %s. Failed to load namespace %s." 
+                            (format "Couldn't load plugin %s. Failed to load namespace %s. This could be caused by a misspelling or a missing dependency." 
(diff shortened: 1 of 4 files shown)


src/kaocha/plugin.clj
```diff
@@ -16,22 +16,48 @@
...
-  (if (qualified-keyword? plugin-name)
-    (when-not (try-require (symbol (str (namespace plugin-name) "." (name plugin-name))))
-      (try-require (symbol (namespace plugin-name))))
-    (try-require (symbol (name plugin-name)))))
+  (let [full-name (symbol (str (namespace plugin-name) "." (name plugin-name)))]
+    (if (qualified-keyword? plugin-name)
+      (if-let [full-result (try-require full-name)]
+        [[full-result full-name]]
+        [[false full-name] (try-require-info (symbol (namespace plugin-name))) ])
+      [(try-require-info (symbol (name plugin-name)))])))
```
plexus, line 28: Could you add a doc string explaining what the inputs/outputs of this function are? I'm having a hard time seeing how to interpret the nested vectors.

src/kaocha/plugin.clj
```diff
@@ -16,22 +16,48 @@
...
 (defmethod -register :default [name plugins]
-  (output/error "Couldn't load plugin " name)
   (throw+ {:kaocha/early-exit 254} nil (str "Couldn't load plugin " name)))
 
 (defn register [plugin-name plugin-stack]
-  (try-load-third-party-lib plugin-name)
-  (-register plugin-name plugin-stack))
+  (let [ns-result (try-load-third-party-lib plugin-name)
+        successful-ns (map second (filter first ns-result))
+        failed-ns (map second (filter (complement first) ns-result))
```
plexus, line 39: Similar comment as above, I'm having a hard time following this because it's not clear what `first` and `second` are. Can we write this in a clearer way?

src/kaocha/plugin.clj
```diff
@@ -16,22 +16,48 @@
...
+        plugin-result (try+ (-register plugin-name plugin-stack)
+                            (catch [:kaocha/early-exit 254] _ false)) ]
```
plexus, line 41: Instead of throwing `:kaocha/early-exit` on line 34 and catching it again, how about we change `-register :default` to throw something like `:kaocha.plugin/not-loaded` or `:kaocha.plugin/unavaiable`, and catch that here? These `early-exit` exceptions are specifically for ending the process with an exit code, I don't think we should start catching these. We should be the only ones calling `-register` so I think it's fine to change its contract.

src/kaocha/plugin.clj
```diff
@@ -16,22 +16,48 @@
...
+        successful-ns (map second (filter first ns-result))
+        failed-ns (map second (filter (complement first) ns-result))
+        plugin-result (try+ (-register plugin-name plugin-stack)
+                            (catch [:kaocha/early-exit 254] _ false)) ] 
+    (cond
+      ;Namespaces succeeded, but plugin itself failed.
+      (and (not plugin-result)
+           (every? true? (map first ns-result))) (output/error-and-throw {:kaocha/early-exit 254} nil
+                                                  (format "Couldn't load plugin %s, but loaded %s." 
+                                                          plugin-name (first successful-ns)))
```
plexus, line 47: I think we can make this a little more clear and explicit.

`Failed loading plugin %s. The plugin was not defined after loading the %s namespace.`

plexus (commented):
Thanks for these changes @alysbrooks! Looking at the tests I can see that it's going to be an improvement, so I'm ok with merging this, but I do find the code quite hard to follow, so I left some comments in case you want to refactor this a bit.

src/kaocha/plugin.clj
alysbrooks: Hmm, I don't know if it's enough to add a docstring and maybe some names via `let` or if I should also change the representation from vectors.

Right now, successfully loading our hook plugin's namespace would be represented as `[true :kaocha.plugins.hooks]` but I could see that being represented as `{:result true :ns :kaocha.plugins.hooks}`. The second would be way nice for further changes, which might be reason enough.

alysbrooks: I held off on the docstring, but I did a small refactoring here because I think it will give us a better sense of whether a slight refactoring and better comments is enough or if the hash-map representation is better.

src/kaocha/plugin.clj
alysbrooks: Changed!

src/kaocha/plugin.clj
alysbrooks: Yeah, I think that's better. I also added more explanation while I was at it.

It now reads `Couldn't load plugin %s. The plugin was not defined after loading namespace %s. Is the file missing a defplugin?`

plexus:
Merged! I changed some of the `;` comments to use `;;` instead, old school LISP commenting conventions dictate that `;` comments are for end-of-line comments that you want to align on the right side next to your code. Because of this Emacs will indent them differently than when using `;;`.

PR #182, opened 2020-12-01
Fix arg order in load-error generator

robhanlon22:
The arguments to ex-info in this generator were transposed. Previously, any time spec tried to generate a `:kaocha.test-plan/load-error`, it threw a ClassCastException.


alysbrooks:
Looks good! 
spongebob_thanks

plexus:
Please don't forget the CHANGELOG :)

PR #187, opened 2020-12-17
Warn when the user has no config

alysbrooks:
If we change the defaults and create a new reader literal for them, `#kaocha/v2`, people with a tests.edn will be fine—they're either using a completely custom config or they're loading a set of defaults they're happy with using `#kaocha/v1`. However, people not using a configuration file at all might see changes in behavior.

To help avoid this, we should warn users when they don't have a configuration file.

```
WARNING: Did not load a configuration file and using the defaults.
This is fine for experimenting, but for long-term use, we recommend creating a configuration file to avoid changes in behavior between releases.
To create a configuration file using the current defaults, create a file named tests.edn that contains 'kaocha/v1 {}'.
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ## Fixed
 
-- Clearly alert the user that Cloure 1.9 isn't supported, rather than
+- Clearly alert the user that Clojure versions before 1.9 aren't supported, rather than
     failing on whatever 1.9 functionality happens to be invoked first.
 - Fixed an issue with the definition of spec `:kaocha.test-plan/load-error` that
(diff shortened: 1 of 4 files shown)


plexus:
Looks good! Just needs a CHANGELOG entry. I was going to say warnings like this should be configurable, but in this case the warning IS configurable, just create a `tests.edn`

plexus:
typo

alysbrooks:
I suppose we could also add a way to say, "yes, I want the latest defaults." A reader literal called `#kaocha/latest`, maybe?

alysbrooks:
Added a `CHANGELOG` entry!

plexus:
puppies

PR #189, opened 2020-12-18
Add .gitignore support to --watch

alysbrooks:
`--watch` lets you ignore certain files using `kaocha.watch/ignore`, which allows you to avoid unnecessary test runs, perhaps because your tests or editor are creating temporary files. However some of these patterns are probably already in `.gitignore` or similar files. 

This PR adds initial support for including ignore patterns from `.gitignore`. To enable it, add `kaocha.watch/use-ignore-file` to `tests.edn`.

Remaining tasks:
- [x] Add support for `.gitignore` files at other points in the hierarchy
- [x] Investigate exactly how git handles escaping trailing spaces
- [x] Add documentation and CHANGELOG entry

Possible tasks:
- [x] Add support for other ignore files with similar syntax
- [ ] Add a command-line argument
- [ ] Warn if watch ignores source or test files
- [x] Always reload .gitignores like we do `tests.edn`. Right now, it will still use the latest `.gitignore` if it's already reloading, but merely changing `.gitignore` doesn't trigger a rerun.


alysbrooks:
Agh, I meant "deterministic," not realistic in the last commit message. Not sure how we feel about rewriting history to fix mistakes like that.

plexus:
If it was already on `main` I would leave it, but since this is "your" branch you're free to `git commit --amend` that. I generally have no problems with amend/rebase/cherry-pick etc. Use them when it makes sense.

If you know others are also working on (or on top of) a branch then you need to weigh the pros and cons, communicate, and coordinate, ideally before and not after changing history.

alysbrooks:
I had already pushed another commit when I saw your message so I'll probably just leave it.

alysbrooks:
Hmm, if we ever want to add `.hgignore` or similar support for another VCS, that technically could cause a breaking change if:
* Someone's global `.hgignore` suddenly takes effect. This is unlikely to be a big issue because files in global ignore files are presumably files the user never modifies directly or cares much about.
* They turned this setting on in a repository with both `.gitignore` and `.hgignore`, although that's a pretty unusual setup. (I've actually done this in order to deploy via Git and version my files using Mercurial.)

We could address this problem by requiring the user to specify the type of ignore files to use. For example, `kaocha.watch/use-ignore-file` would be set to `:git` (for `.gitignore`), `:ignore` (for .ignore), or `:all`.

Ultimately, I don't think we have to worry about this because the first situation doesn't seem very problematic and the second situation is pretty unlikely, so this comment is mostly to record my thought process.

src/kaocha/watch.clj
```diff
@@ -86,6 +87,81 @@
         patterns (map #(.getPathMatcher fs (str "glob:" %)) patterns)]
     (some #(.matches % path) patterns)))
 
+
+(defn convert 
+  "Converts a Git-style ignore pattern into the equivalent pattern that Java PathMatcher uses."
+  [pattern] 
+  (let [cleaned  (-> pattern 
+                   ;If a Git pattern has a trailing space, it should be stripped (unless escaped)
+                   ;Example: 'src/test ' => 'src/test'
```
plexus, line 96: Please use `;;` instead of `;`

src/kaocha/watch.clj
```diff
@@ -86,6 +87,81 @@
...
+                   ; (re-find #"[{}]" pattern) (str/replace pattern #"\{(.*)\}" "\\\\{$1\\\\}"  ) 
+                   (str/replace #"\{(.*)\}" "\\\\{$1\\\\}"))]
+        (cond 
+                    ;If it starts with a single *, it should have **
+                    ;Example: *.html => **.html
+                    (re-find #"^\*[^*]" cleaned) (str \* cleaned)
+
+                    ;If a Git pattern ends with a slash, that represents everything underneath
+                    ;Example: src/ => src/**
+                    (re-find #"/$" cleaned) (str cleaned "**")
```
plexus, line 127: Are these two mutually exclusive, or is it possible to start with `**` and end with `/`? (if so `cond->` might be useful)

src/kaocha/watch.clj
```diff
@@ -86,6 +87,81 @@
...
+
+(defn convert 
+  "Converts a Git-style ignore pattern into the equivalent pattern that Java PathMatcher uses."
+  [pattern] 
+  (let [cleaned  (-> pattern 
+                   ;If a Git pattern has a trailing space, it should be stripped (unless escaped)
+                   ;Example: 'src/test ' => 'src/test'
+                   (str/replace #"([^\\]) +$" "$1")
+
+                   ;If a Git paattern contains a double star bordering no path
```
plexus, line 99: `paattern` typo

plexus (approved):
Looks great! I left a few small comments. Regarding `;;` vs `;` this seems to come up more and more often with different editors taking different defaults, I commented on it some more here https://clojurians-log.clojureverse.org/calva/2021-01-20/1611164080.197800

tl;dr "not a hill I'm willing to die on, and I may end up having to suck it up, but for now I will continue to be petty about it"

src/kaocha/watch.clj
alysbrooks: My current commenting plugin (commentary) doesn't distinguish between whole-line and part-of-line comments. Happily, there appears to be another solid one that does (tcomment), so I'll try that one.

src/kaocha/watch.clj
alysbrooks: The documentation isn't really clear, but after testing with git, looks like they aren't.

PR #191, opened 2021-01-18
Balance braces in example tests.edn in README

socksy:
Fix the tests.edn in the README. Small change, but lets someone copy and paste it directly if they want.


plexus:
Vielen Dank!

puppies

Issue #184, opened 2020-12-10
Document meta-merge behavior of config


plexus:
When processing the `#kaocha/v1` reader macro we merge the user's config into the default config using [meta-merge](https://github.com/weavejester/meta-merge). This means that you can add `^:prepend`, `:^append`, and `^:replace` metadata to influence how the merging should happen.

For example, user-configured plugins are added onto the list of default plugins `[:kaocha.plugin/randomize :kaocha.plugin/filter :kaocha.plugin/capture-output]`, rather than replacing them. If you want to use a different variant of the filter plugin then you could use `:plugins ^:replace [...]` instead.

We do mention `meta-merge` in the docs, but only when used explicitly with a `#meta-merge` reader tag. There's also one example of a `^:replace` in the cucumber test/doc for filter, but presented without much comment.

PR #192, opened 2021-01-25
Initial draft of material about meta-merge.

alysbrooks:
Adds some documentation about meta-merge, including some examples. I'm not 100 percent sure about what is automatically meta-merged, hence the draft.

Addresses  #184.

Questions:
- [x] ~~What parts of the configuration already have `meta-merge` applied? I think it's "everything except for select forms"~~ Kind of? We apply `meta-merge` to everything, and use `^:replace` for certain things.
- [x] ~~Can `^:prepend`, :`^append`, and` ^:replace` be used with `:kaocha/tests`? I can't get it to work.~~ They were supposed to, but there was a bug.

### DIFF
--- a/doc/03_configuration.md
+++ b/doc/03_configuration.md
@@ -94,9 +94,10 @@ The primary advantage to using the `#meta-merge` tag over aero's
 `#merge` is that meta-merge will do a deep merge of the tree; meaning
 developers can more easily inherit changes to the base configuration.
 
-Meta-merge is also applied by default for ???. You can use `^:prepend`, :`^append`, and` ^:replace` to
-control the merging.  For example, if you wanted to overide the list of default
-plugins rather than add on to it, write:
+Meta-merge is also applied by default for most keys, with some exceptions listed
+below. You can use `^:prepend`, :`^append`, and` ^:replace` to control the
+merging.  For example, if you wanted to overide the list of default plugins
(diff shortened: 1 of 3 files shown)


src/kaocha/config.clj
```diff
@@ -19,12 +19,15 @@
 
 (defn replace-by-default [config k]
   (if-let [v (get config k)]
-    (if (#{:prepend :append} (meta v))
+    (if (some #{:prepend :append} (keys (meta v)))
```
plexus, line 22: :+1: had to stare at this for a few seconds before realizing the old version is borked.

doc/03_configuration.md
```diff
@@ -94,6 +94,24 @@ The primary advantage to using the `#meta-merge` tag over aero's
 `#merge` is that meta-merge will do a deep merge of the tree; meaning
 developers can more easily inherit changes to the base configuration.
 
+Meta-merge is also applied by default for ???. You can use `^:prepend`, :`^append`, and` ^:replace` to
```
plexus, line 97: Still a `???` that I think needs filling in. Apart from that it looks good.

alysbrooks: I switched to a computer that didn't have the latest changes, rebased, and then force-pushed, meaning the changes were lost from GitHub. I restored them from another copy.

Issue #190, opened 2021-01-08
watch-test doesn't seem to actually be testing the behavior


alysbrooks:
I was trying to write a similar test for the `.gitignore` feature, and I wasn't sure it was actually testing so I tried commenting things out to get it to fail, without success. It seems like the same is true for the test I based it off of `kaocha.watch-test/watch-test`

It looks like this is using `is` with a value as its first argument:

```clojure
(is (str/replace "[(F)]\n\nFAIL in foo.bar-test/xxx-test (bar_test.clj:1)\nExpected:\n  :xxx\nActual:\n  -:xxx +:yyy\n1 tests, 1 assertions, 1 failures.\n\n[watch] Reloading #{foo.bar-test}\n[watch] Re-running failed tests #{:foo.bar-test/xxx-test}\n[(F)]\n\nFAIL in foo.bar-test/xxx-test (bar_test.clj:1)\nExpected:\n  :xxx\nActual:\n  -:xxx +:zzz\n1 tests, 1 assertions, 1 failures.\n\n[watch] watching stopped.\n"
                     "foo"
                     prefix)
        @out-str))
```

When I change it to the following, it fails:

```clojure
(is (= (str/replace "[(F)]\n\nFAIL in foo.bar-test/xxx-test (bar_test.clj:1)\nExpected:\n  :xxx\nActual:\n  -:xxx +:yyy\n1 tests, 1 assertions, 1 failures.\n\n[watch] Reloading #{foo.bar-test}\n[watch] Re-running failed tests #{:foo.bar-test/xxx-test}\n[(F)]\n\nFAIL in foo.bar-test/xxx-test (bar_test.clj:1)\nExpected:\n  :xxx\nActual:\n  -:xxx +:zzz\n1 tests, 1 assertions, 1 failures.\n\n[watch] watching stopped.\n"
                     "foo"
                     prefix)
        @out-str)))
```

Additionally, commenting out some forms in the middle of the test doesn't appear to change the result.

plexus:
Yup, good find, that test is not testing anything. Apparently I wrote it two years ago on the morning of my birthday, I guess I wasn't very awake yet...

plexus:
No need dragging a useless test around. If it's not immediately obvious how to fix it then I think it's better if we just delete it, or at least give it a `^:kaocha/pending`

dharrigan:
What would you like to do? (I'm in that area right now - remove it, or mark it as pending).

plexus:
[...]

PR #193, opened 2021-01-27
Provide a warning when no tests are found.

### LINKED ISSUE #34: Provide a warning when no tests are found

dharrigan:
Fixes #34

-=david=-

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - `kaocha.plugin.alpha.spec-test-check` now honors command line arguments based
   upon all of the configured STC suites rather than the static
   `:generative-fdef-checks` selector.
+- Provide a warning when no tests are found.
 
 ## Changed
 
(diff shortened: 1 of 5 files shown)


alysbrooks:
Overall, looks good.

Can you rebase this?

I'm not sure what the best way forward on the broken test is, as it is already broken (#190). I would suggest borrowing code from `watch-test` so it has at least one test. In any case, I don't think you need to wrap `try+` in another `try`.

dharrigan:
Hi,

Thank you for your kind feedback. I've rebased my branch against master, all is up-to-date. I've removed the usage of the `try+` being wrapped with a `try` (although, it does exist in one other place, namelyy `test/unit/kaocha/report_test.clj`.

Waiting on some feedback from #190 to decide upon what to do with the watch test :-)

dharrigan:
Updated ;)

alysbrooks:
Thanks! Can you add a CHANGELOG entry for this fix? Also, it looks like you added extra lines to `/report_test.clj`

alysbrooks:
Thanks! 
giphy

dharrigan:
You're most welcome! Maybe I can do more! :)

alysbrooks:
That'd be great! Let me know if you have any questions!

PR #194, opened 2021-01-28
Update test deps

alysbrooks:
Updates some dependencies. We probably want to be more careful about updating non-test dependencies, so I'm leaving those off for now.


alysbrooks:
I realized that updating matcher-combinators would potentially be a breaking change for people depending on a 1.* version, so I reverted that.

plexus:
I wouldn't be too hesitant in just upgrading dependencies, that way at leasy we'll find out if upstream broke something (which is rare in clojure land anyway). If the tests pass -> ship-it, if people then report issues we can look into what to do, downgrade, fix it, etc. And meanwhile people can lock to a specific version in their own dependency declaration.

For matcher-combinators it might be cool to look into making it an optional (BYO) dependency, I think that's quite doable. I do think everyone should be using it, but if they don't then we shouldn't force it to be part of their project.

plexus:
cat-orb

Almost forgot the traditional PR-merged-gif :)

PR #196, opened 2021-02-12
Document the skip-add-classpath? flag

plexus:

### DIFF
--- a/doc/03_configuration.md
+++ b/doc/03_configuration.md
@@ -155,11 +155,13 @@ duplicated between `tests.edn` and your project configuration (`deps.edn`,
 However, often people will have these paths in their `deps.edn` anyway (e.g. in
 a `:test` or `:dev` profile), so they are accessible for interactive evaluation
 (REPL, in-buffer eval). In some cases you also do not *want* Kaocha to add
-these, because the way we make sure we have a dynamic classloader so we can add
-entries at runtime can confuse third party tools.
+these, because it can confuse third-party tools. This is because Kaocha needs a
+`DynamicClassLoader` to be able to add classpath entries at runtime. If the
+current classloader is not a `DynamicClassLoader` then we set one up ourselves,
+but there have been cases where this caused issues with other tools.
 
 In these cases you can add `:kaocha.testable/skip-add-classpath? false` to the
-test suite definition, to disable this behavior.
+test suite definition to disable this behavior.
 
 ### :kaocha.type/clojure.test
 


doc/03_configuration.md
```diff
@@ -145,6 +145,22 @@ This is what the `:unit` suite looks like after expansion:
...
+Kaocha will add any entries in `:kaocha/test-paths` to the classpath, if they
+aren't on there already. This is done so this information doesn't need to be
+duplicated between `tests.edn` and your project configuration (`deps.edn`,
+`project.clj`, etc).
+
+However, often people will have these paths in their `deps.edn` anyway (e.g. in
+a `:test` or `:dev` profile), so they are accessible for interactive evaluation
+(REPL, in-buffer eval). In some cases you also do not *want* Kaocha to add
+these, because the way we make sure we have a dynamic classloader so we can add
+entries at runtime can confuse third party tools.
```
alysbrooks, line 159: "the way we make sure we have a dynamic classloader so we can add
entries at runtime" is a little wordy (the phrasing isn't bad in itself, it's just long in this context).

Maybe something like "In some cases you also do not *want* Kaocha to add
these, because the automatic classpath handling can confuse third-party tools. It's specifically the way Kaocha ensures we have a dynamic classloader to add these entries dynamically that confuses them." 

That second sentence is probably not necessary. If it is, I think you need more information so it's clear why it's necessary to the reader.

doc/03_configuration.md
```diff
@@ -145,6 +145,22 @@ This is what the `:unit` suite looks like after expansion:
...
+
+In these cases you can add `:kaocha.testable/skip-add-classpath? false` to the
+test suite definition, to disable this behavior.
```
alysbrooks, line 162: I don't think you need this comma.

alysbrooks (commented):
I think this PR is ready to merge after these changes.

doc/03_configuration.md
```diff
@@ -155,11 +155,13 @@ duplicated between `tests.edn` and your project configuration (`deps.edn`,
...
 a `:test` or `:dev` profile), so they are accessible for interactive evaluation
 (REPL, in-buffer eval). In some cases you also do not *want* Kaocha to add
-these, because the way we make sure we have a dynamic classloader so we can add
-entries at runtime can confuse third party tools.
+these, because it can confuse third-party tools. This is because Kaocha needs a
+`DynamicClassLoader` to be able to add classpath entries at runtime. If the
+current classloader is not a `DynamicClassLoader` then we set one up ourselves,
+but there have been cases where this caused issues with other tools.
 
 In these cases you can add `:kaocha.testable/skip-add-classpath? false` to the
```
arichiardi, line 163: Hi there, I was wondering if we want `:kaocha.testable/skip-add-classpath? true` instead here cause that would actually skip the `add-classpath` statement in the code?

If not apologies, maybe I am not understanding too much the use case in the README...

plexus: Uh yes I believe you're right

arichiardi: Ok no worries I can open a PR if that's the case.

Issue #185, opened 2020-12-12
Make deep-diff configurable


alysbrooks:
As helpful as deep-diff is, it may not be your preference or you may want to turn it off temporarily in order to more easily copy the actual input (as opposed to a diff). This is possible already, but it's not very straightforward—you have to create a different `kaocha.report/fail-summary` method. I think you can technically redefine `kaocha.report/fail-summary` in your own code, but I'm not sure this is the right approach.

A way to redefine this on the commandline would be great, to support switching deep-diff off for  one-off case like #80. I'm not sure if we can pass configuration options to reporters. If so, we'd just have to update the reporters accordingly. From the docs, it sounds like we could also create a different fail-summary function that users could optionally turn on.

plexus:
To elaborate a bit more here, the `fail-summary` is part of the `kaocha.report/result` reporter, a single-purpose reporter which waits for the `:summary` event at the end of the test run, and then prints out all failures/errors. This is kaocha-specific behavior, in plain clojure.test these error/fail messages are printed immediately, instead of as an aggregated report at the end of the run.

This `summary` reporter is combined with another reporter which does the progress reporting during the test run.

``` clojure
(def dots
  "Reporter that prints progress as a sequence of dots and letters."
  [dots* result])

(def documentation
  "Reporter that prints an overview of all tests bein run using indentation."
  [doc result])
```

So here we have lower level `dots*` and `doc` reporters which handle all `:pass`/`:error`/`:fail`/start/end etc messages, combined with the `result` reporter which prints out the details at the end. With what we currently have the best approach to change the summary behavior by an end user is to create their own version of this `result` reporter, and then combine it in `tests.edn` with `dots*`/`doc`/`progress`.

``` clojure
:reporter [kaocha.report/doc kaocha.report/result]
```

`fail-summary` itself relies on `kaocha.report/print-expr`, a multimethod which prints out some formatted version of whatever is inside the `(is ...)`, dispatching on the predicate. So `(is (= ...))` forms are handled like this:

``` clojure
(defmethod print-expr '= [m] ;; <- m here is the clojure.test event map with :expected, :actual, :message, :type
  ,,,)
```

So this could also be overridden in user space.

``` clojure
(require 'kaocha.report)
(defmethod kaocha.report/print-expr '= [m]
  (println "expected:" (pr-str (:expected m)))
  (println "  actual:" (pr-str (:actual m))))
```

[...]

PR #197, opened 2021-02-19
Document customizations of fail-summary

alysbrooks:
First PR to address #185 . This PR adds documentation about the existing options for configuring how results are reported, including whether `deep-diff` is used.


plexus:
Looks good to me! Great idea to include an image here to point at the different parts.

Issue #198, opened 2021-02-19
Notifier plugin is broken on Windows


ArielA147:
Notifier plugin throws an exception if `which` doesn't exist over windows.

### Exception message:
```
; Execution error (IOException) at java.lang.ProcessImpl/create (REPL:-2).
; CreateProcess error=2, The system cannot find the file specified
```
### Stacktrace:
```
java.lang.ProcessBuilder/start (:-1)
clojure.java.shell/sh (shell.clj:113)
clojure.java.shell/sh (shell.clj:79)
kaocha.plugin.notifier/exists? (notifier.clj:16)
kaocha.plugin.notifier/exists? (notifier.clj:15)
kaocha.plugin.notifier/detect-command (notifier.clj:20)
kaocha.plugin.notifier/detect-command (notifier.clj:18)
kaocha.plugin.notifier/notifier-config-hook (notifier.clj:105)
kaocha.plugin.notifier/notifier-config-hook (notifier.clj:91)
clojure.core/apply (core.clj:667)
clojure.core/apply (core.clj:660)
kaocha.plugin/run-hook* (plugin.clj:50)
clojure.lang.PersistentVector/reduce (PersistentVector.java:343)
clojure.core/reduce (core.clj:6827)
clojure.core/reduce (core.clj:6810)
kaocha.plugin/run-hook* (plugin.clj:48)
kaocha.plugin/run-hook* (plugin.clj:47)
clojure.core/apply (core.clj:671)
clojure.core/apply (core.clj:660)
kaocha.plugin/run-hook (plugin.clj:59)
kaocha.plugin/run-hook (plugin.clj:58)
kaocha.repl/config (repl.clj:88)
kaocha.repl/config (repl.clj:67)
kaocha.repl/run (repl.clj:151)
kaocha.repl/run (repl.clj:124)
clojurians-log.slack-mgs-test/eval35241 (NO_SOURCE_FILE:76)
clojure.lang.Compiler/eval (Compiler.java:7177)
clojure.core/eval (core.clj:3214)
clojure.core/eval (core.clj:3210)
nrepl.middleware.interruptible-eval/evaluate (interruptible_eval.clj:87)
clojure.core/apply (core.clj:665)
clojure.core/with-bindings* (core.clj:1973)
nrepl.middleware.interruptible-eval/evaluate (interruptible_eval.clj:87)
clojure.main/repl (main.clj:437)
clojure.main/repl (main.clj:458)
clojure.main/repl (main.clj:368)
[...]

PR #199, opened 2021-02-23
Extend exists? to not cause an issue on Windows.

alysbrooks:
On Windows, there's no `which` command, so our attempt to determine if commands exist fails with an unhelpful error. Addresses #198.

### DIFF
--- a/src/kaocha/plugin/notifier.clj
+++ b/src/kaocha/plugin/notifier.clj
@@ -6,8 +6,10 @@
             [kaocha.result :as result]
             [kaocha.shellwords :refer [shellwords]]
             [clojure.string :as str]
-            [clojure.java.io :as io])
-  (:import [java.nio.file Files]))
+            [clojure.java.io :as io]
+            [slingshot.slingshot :refer [throw+]])
+  (:import [java.nio.file Files]
+           [java.io IOException] ))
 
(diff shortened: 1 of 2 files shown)


src/kaocha/plugin/notifier.clj
```diff
@@ -6,14 +6,22 @@
...
 
 (defn exists? [program]
-  (= 0 (:exit (sh "which" program))))
+  (let [cmd (if 
+              (re-find #"Windows" (System/getProperty "os.name"))
+              "where.exe" "which" )]
+    (try 
+      (= 0 (:exit (sh cmd program)))
+      (catch IOException e  ;in the unlikely event where.exe or which isn't available
+        (throw+ {:kaocha/early-exit 0} e)))))
```
alysbrooks, line 24: I'm not aware of any platforms that wouldn't be covered by this logic, so I could see an argument for this code is unnecessary. On the other hand, I think shelling out has an inherent risk, particularly in the context of something running on a VM, so maybe all calls to sh should have a similar guard?

plexus: I agree, shelling out can always error for whatever reason. But maybe throwing a `:kaocha/early-exit` is a bit much here. We can still run tests without notifications, so maybe just outputting a warning is better.

PR #202, opened 2021-03-03
Lower num-tests and see if that helps.

alysbrooks:
`kaocha.plugin.alpha.spec-test-check` consistently fails with what we think is an out-of-memory error. It might be that the generative tests are creating too much garbage, so doing fewer might help.


alysbrooks:
I marked this as a draft because I'll likely make a clean version of this as a separate PR (or possibly commit it directly to main, it's a small change).

alysbrooks:
Apparently setting it to 50 makes it work, so I've committed that change directly to main.

PR #203, opened 2021-03-04
Add support for Java's notification messages.

alysbrooks:
The JVM has built-in functionality for notification messages. We don't want to replace the existing use of `notify-send` and `terminal-notifier` because the JVM notifications don't look great on every platform, but it works as a fallback.

108773981-b9b74400-7524-11eb-8eab-bf677c990f61
108773922-a5734700-7524-11eb-8e63-66a89fa9e891

### DIFF
--- a/src/kaocha/plugin/notifier.clj
+++ b/src/kaocha/plugin/notifier.clj
@@ -82,9 +82,15 @@
   Not preferred over shelling out because the built-in notification sometimes
   looks out of place, and isn't consistently available on Linux."
   [result]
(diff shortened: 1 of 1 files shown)


alysbrooks:
Here's what it looks like on Ubuntu with GNOME:
Screenshot from 2021-03-04 15-35-41
Screenshot from 2021-03-04 15-36-16

plexus:
Looks good. It seems on windows we could shell out to PowerShell for this, but let's chalk that up as a future improvement. (Windows apparently calls these "Toaster notifications", so you know what to google for.)

PR #218, opened 2021-04-09
Configurable deep diff

alysbrooks:
Addresses #185. 

To quote the new docs,
>Kaocha uses `deep-diff2` when tests fail to distinguish the difference between
the actual and expected values. If you don't like the format, or if it provides
unhelpful output in a particular scenario, you can turn it off using the
`--diff-style :none` option.'

I'm open to other names than `diff-style`. It's what @plexus suggested, and I think it's pretty good. Ideally, it would indicate that it's the diff *used for errors*, though.


src/kaocha/config.clj
```diff
@@ -115,12 +115,14 @@
...
-    (:watch options)             (assoc :kaocha/watch? (:watch options))
-    (some? (:color options))     (assoc :kaocha/color? (:color options))
-    (:plugin options)            (update :kaocha/plugins #(distinct (concat % (:plugin options))))
-    true                         (assoc :kaocha/cli-options options)))
+    (some? (:fail-fast options))  (assoc :kaocha/fail-fast? true)
+    (:reporter options)           (assoc :kaocha/reporter (:reporter options))
+    (:watch options)              (assoc :kaocha/watch? (:watch options))
+    (some? (:color options))      (assoc :kaocha/color? (:color options))
+    ;; (some? (:diff-style options)) (assoc :kaocha/diff-style (:diff-style options))
+    (some? (:diff-style options)) (assoc :diff-style (:diff-style options))
```
plexus, line 123: Please use namespaced keywords in the config map.

```
    (some? (:diff-style options)) (assoc :kaocha/diff-style (:diff-style options))
```

plexus:
looks good to me!

Issue #208, opened 2021-03-17
Clojure syntax error exception swallowed?


lread:
**First and foremost**: thanks so much for kaocha, I love it!

After upgrading from kaocha `1.0.732` to `1.0.829` I noticed that a Clojure syntax error in one of my tests was not being reported by kaocha.  

**Reproduction**
`deps.edn`
```Clojure
{:deps {org.clojure/clojure {:mvn/version "1.10.3"}}
 :aliases {:kaocha {:extra-deps {lambdaisland/kaocha {:mvn/version "1.0.829"}}
                    :main-opts ["-m" "kaocha.runner"]}}}
```

`test/my_test.clj`
```Clojure
(ns my-test
  (:require [clojure.test :refer [deftest is]]))

oopsie

(deftest testing
  (is (= 1 1)))
```

**Expected Behaviour**
I would expect that kaocha would convey that I made an `oopsie` in my test and return a non-zero exit code.

**Actual Behaviour**
Kaocha is not showing me my error and returning an exit code of `0`.

```Shell
> clojure -M:kaocha
WARNING: Did not load a configuration file and using the defaults.
This is fine for experimenting, but for long-term use, we recommend creating a configuration file to avoid changes in behavior between releases.
To create a configuration file using the current defaults, create a file named tests.edn that contains 'kaocha/v1 {}'.
WARNING: No tests were found, make sure :test-paths and :ns-patterns are configured correctly in tests.edn.
> echo $?
0
```

**Behaviour under v1.0.732**
If we downgrade kaocha:
`deps.edn`
```Clojure
{:deps {org.clojure/clojure {:mvn/version "1.10.3"}}
 :aliases {:kaocha {:extra-deps {lambdaisland/kaocha {:mvn/version "1.0.732"}}
                    :main-opts ["-m" "kaocha.runner"]}}}
```

We see the syntax error and get our non-zero exit code:
```Shell
> clojure -M:kaocha
[E]
Randomized with --seed 1554809814

ERROR in unit (my_test.clj:0)
Exception: clojure.lang.Compiler$CompilerException: Syntax error compiling at (my_test.clj:0:0).
#:clojure.error{:phase :compile-syntax-check, :line 0, :column 0, :source "my_test.clj"}
 at clojure.lang.Compiler.analyze (Compiler.java:6812)
    ...
    kaocha.ns$required_ns.invokeStatic (ns.clj:13)
    kaocha.ns$required_ns.invoke (ns.clj:11)
[...]

PR #222, opened 2021-04-29
Don't let the "no tests found" message shadow the load error

plexus:
When files fail to parse then it's likely no tests can be loaded, in that case
we need to show the error, not the generic "check your config" message.

This should go a long way towards addressing #208, although there still may be other cases where the load error doesn't properly come through.

### DIFF
--- a/src/kaocha/api.clj
+++ b/src/kaocha/api.clj
@@ -98,7 +98,9 @@
           (with-bindings (config/binding-map config)
             (let [config (resolve-reporter config)]
               (let [test-plan (test-plan config)]
-                (when-not (some hierarchy/leaf? (testable/test-seq test-plan))
+                (when-not (some #(or (hierarchy/leaf? %)
+                                     (::testable/load-error %))
+                                (testable/test-seq test-plan))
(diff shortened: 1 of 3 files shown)


alysbrooks:
This looks good to merge to me!

plexus:
Then let's do it!

plexus:
dinner_party

src/kaocha/api.clj
```diff
@@ -98,7 +98,9 @@
           (with-bindings (config/binding-map config)
             (let [config (resolve-reporter config)]
               (let [test-plan (test-plan config)]
-                (when-not (some hierarchy/leaf? (testable/test-seq test-plan))
+                (when-not (some #(or (hierarchy/leaf? %)
+                                     (::testable/load-error %))
+                                (testable/test-seq test-plan))
                   (output/warn (str "No tests were found, make sure :test-paths and "
                                     ":ns-patterns are configured correctly in tests.edn."))
                   (throw+ {:kaocha/early-exit 0}))
```
arichiardi, line 106: I wonder if this one carries the original compilation error with it - see my comment in #208

PR #224, opened 2021-05-21
Add Windows CI

alysbrooks:
This adds Windows to our CI configuration. Once this is figured out, it can serve as a template for other projects. We still have some outstanding issues with these tests, so I'll either wait to merge this or skip the problematic tests on just our Windows build.

- [x] Confirm tests actually run.
- [x] Figure out why errors from `clojure` aren't being treated as failures.
- [ ] Decide whether to skip problematic tests, and if so, add metadata for tests that don't work on Windows.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -44,7 +44,7 @@ jobs:
           command: clojure -h
           shell: powershell.exe
       - run:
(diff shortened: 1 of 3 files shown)


alysbrooks:
While removing some platform-specific code was valuable, I don't think troubleshooting this kind of extended testing is one of our priorities for the moment, so I'm keeping the changes but disabling Windows tests.

PR #227, opened 2021-06-07
Fix print-invocations for boolean flags

emlyn:
If you run `./bin/kaocha --no-capture-output` and a test fails, it suggests using `bin/kaocha --capture-output false --focus '...'`, which if you try to run it fails with `No such suite: :false, valid options: :unit.` since the `--capture-output` option does not expect any parameters.
To fix this, I added handling for boolean options so that they should get printed correctly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Added
 
 ## Fixed
+- Fix printing of boolean options in the print-invocations plugin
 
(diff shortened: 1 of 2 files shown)


alysbrooks:
Thank you for the fix! If you have time, can you add a quick comment to the CHANGELOG?

emlyn:
Sure, done!

PR #233, opened 2021-07-20
Fix and clean up load-error handling, remove Orchestra

plexus:
This makes kaocha.watch use the mechanisms already available in
`kaocha.testable` to deal with and report load errors. It also does a general
pass of cleaning things up, because we had created a bit of a mess.

`handle-load-error` is gone, the testable implementations should no longer need
this, just let kaocha.testable handle this stuff for you.

It also seems at some point we went from `:kaocha.test-plan/load-error` to
`:kaocha.testable/load-error`, and ended up with a bit of a hodge-podge. It's
surprising things still worked as well as they did. I changed the remaining
test-plan/load-error references to testable/load-error, since that now
dominates, and is IMO also the more intuitive, since we apply these on
individual testables (typically the suite).

I removed the Orchestra instrumentation, and disabled the fdef check plugin. These things get in the way too easily, at this point they are causing us a lot of work without adding much value. I also dropped the (outdated) Orchestra dependency. The orchestra plugin is still there, but going forward people should bring their own Orchestra if they still want to use it.

Other small bits
- fixed a reflection warning in the notifier plugin
- version bumps
- added portal (dev use)
- somewhat more comprehensive error message when the reporter throws (luckily a rare occasion, unless orchestra starts feeding it random bullshit)
- lots of whitespace fixes

Closes #226

PR #234, opened 2021-07-21
Run tests in parallel

alysbrooks:
Very much a work in progress! Mostly opening this PR so I have a place to take and share notes.

- [x] Ensure #233 gets merged
- [ ] ~~Add some way of monitoring for debug or profiling?~~
- [x] ~~Investigate test-level parallelization~~ Add configuration for test-level parallelization
- [x] Test on several real-world test suites
   - [x] Regal
   - [x] deep-diff2
   - [x] deja-fu
- [ ] ~~Ensure test reporters print meaningful results~~
   - [ ] ~~Dots~~
   - [ ] ~~Documentation~~
- [ ] Fix test
- [x] Disable based on test type, not just parent type
- [x] Add docs
- [ ] Make sure addition of volatile in `kaocha.watch` doesn't interfere with this. See #371.

### DIFF
--- a/repl_sessions/benchmark.clj
+++ b/repl_sessions/benchmark.clj
@@ -1,0 +1,97 @@
+
+(ns benchmark
+  (:require [criterium.core :as c])
+  (:import [java.util.concurrent Executors ])
+  )
+
+(def thread-pool (Executors/newFixedThreadPool 10))
+
+(defn math-direct []
+  (+ 1 1))
+
+(defn math-future []
+  (deref 
+    (future (+ 1 1))))
+
+(defn math-thread []
+  (let [result (atom nil)]
+    (doto (Thread. (fn [] (reset! result (+ 1 1))))
+      (.start)
+      (.join))
+    @result))
+
+(defn math-threadpool []
+  (let [result (atom nil)]
+    (.get (.submit thread-pool (fn [] (reset! result (+ 1 1))) ))
+    @result))
+
+(defn math-threadpool-no-atom []
+  (.get (.submit thread-pool (fn [] (+ 1 1)) )))
+
+
+(c/bench (math-direct) )
+; (out) Evaluation count : 6215391600 in 60 samples of 103589860 calls.
+; (out)              Execution time mean : 2,015262 ns
+; (out)     Execution time std-deviation : 0,497743 ns
+; (out)    Execution time lower quantile : 1,442374 ns ( 2,5%)
+; (out)    Execution time upper quantile : 3,392990 ns (97,5%)
+; (out)                    Overhead used : 7,915626 ns
+; (out) 
+; (out) Found 5 outliers in 60 samples (8,3333 %)
+; (out) 	low-severe	 3 (5,0000 %)
+; (out) 	low-mild	 2 (3,3333 %)
+; (out)  Variance from outliers : 94,6147 % Variance is severely inflated by outliers
+
+(c/bench (math-future) )
+; (out) Evaluation count : 3735420 in 60 samples of 62257 calls.
+; (out)              Execution time mean : 16,635809 µs
+; (out)     Execution time std-deviation : 1,104338 µs
+; (out)    Execution time lower quantile : 15,397518 µs ( 2,5%)
+; (out)    Execution time upper quantile : 19,751883 µs (97,5%)
+; (out)                    Overhead used : 7,915626 ns
+; (out) 
+; (out) Found 6 outliers in 60 samples (10,0000 %)
+; (out) 	low-severe	 3 (5,0000 %)
+; (out) 	low-mild	 3 (5,0000 %)
+; (out)  Variance from outliers : 50,0892 % Variance is severely inflated by outliers
+
+(c/bench (math-thread))
+
+; (out) Evaluation count : 774420 in 60 samples of 12907 calls.
+; (out)              Execution time mean : 82,513236 µs
+; (out)     Execution time std-deviation : 5,706987 µs
+; (out)    Execution time lower quantile : 75,772237 µs ( 2,5%)
+; (out)    Execution time upper quantile : 91,971212 µs (97,5%)
+; (out)                    Overhead used : 7,915626 ns
+; (out) 
+; (out) Found 1 outliers in 60 samples (1,6667 %)
+; (out) 	low-severe	 1 (1,6667 %)
+; (out)  Variance from outliers : 51,7849 % Variance is severely inflated by outliers
+
+(c/bench (math-threadpool))
+; (out) Evaluation count : 3815100 in 60 samples of 63585 calls.
+; (out)              Execution time mean : 16,910124 µs
+; (out)     Execution time std-deviation : 2,443261 µs
+; (out)    Execution time lower quantile : 14,670118 µs ( 2,5%)
+; (out)    Execution time upper quantile : 23,743868 µs (97,5%)
+; (out)                    Overhead used : 7,915626 ns
+; (out) 
+; (out) Found 3 outliers in 60 samples (5,0000 %)
+; (out) 	low-severe	 2 (3,3333 %)
+; (out) 	low-mild	 1 (1,6667 %)
+; (out)  Variance from outliers : 82,4670 % Variance is severely inflated by outliers
+
+
+(c/bench (math-threadpool-no-atom))
+
+; (out) Evaluation count : 3794940 in 60 samples of 63249 calls.
+; (out)              Execution time mean : 16,182655 µs
+; (out)     Execution time std-deviation : 1,215451 µs
+; (out)    Execution time lower quantile : 14,729393 µs ( 2,5%)
+; (out)    Execution time upper quantile : 18,549902 µs (97,5%)
+; (out)                    Overhead used : 7,915626 ns
+; (out) 
+; (out) Found 3 outliers in 60 samples (5,0000 %)
+; (out) 	low-severe	 2 (3,3333 %)
+; (out) 	low-mild	 1 (1,6667 %)
+; (out)  Variance from outliers : 56,7625 % Variance is severely inflated by outliers
--- a/src/kaocha/api.clj
+++ b/src/kaocha/api.clj
@@ -123,11 +123,6 @@
                                       " To investigate, check the :test-paths and :ns-patterns keys in tests.edn.")))
                   (throw+ {:kaocha/early-exit 0}))
 
-                (when (:parallel config)
-                  (output/warn (str "Parallelization enabled. This feature is in "
-                                    "beta If you encounter errors, try "
-                                    "running with the feature disabled and "
-                                    "consider filing an issue.")))
                 (when (find-ns 'matcher-combinators.core)
                   (require 'kaocha.matcher-combinators))
 
@@ -142,8 +137,7 @@
                                       ;; don't know where in the process we've
                                       ;; been interrupted, output capturing may
                                       ;; still be in effect.
-                                      (System/setOut
-                                       orig-out)
+                                      (System/setOut orig-out)
                                       (System/setErr
                                        orig-err)
                                       (binding [history/*history* history]
@@ -160,13 +154,13 @@
                                             on-exit)
                         (let [test-plan (plugin/run-hook :kaocha.hooks/pre-run test-plan)]
                           (binding [testable/*test-plan* test-plan]
-                            (let [test-plan-tests (:kaocha.test-plan/tests test-plan)
-                                  result-tests    (testable/run-testables test-plan-tests test-plan)
-                                  result          (plugin/run-hook :kaocha.hooks/post-run
-                                                                   (-> test-plan
-                                                                       (dissoc :kaocha.test-plan/tests)
-                                                                       (assoc :kaocha.result/tests result-tests)))]
-                              (assert (= (count test-plan-tests) (count (:kaocha.result/tests result))))
+                            (let [result-tests (testable/run-testables-parent test-plan test-plan)
+                                  result       (plugin/run-hook :kaocha.hooks/post-run
+                                                          (-> test-plan
+                                                              (dissoc :kaocha.test-plan/tests)
+                                                              (assoc :kaocha.result/tests result-tests)))]
+                              (assert (= (count (:kaocha.test-plan/tests test-plan))
+                                         (count (:kaocha.result/tests result))))
                               (-> result
                                   result/testable-totals
                                   result/totals->clojure-test-summary
--- a/src/kaocha/api.clj
+++ b/src/kaocha/api.clj
@@ -124,10 +124,10 @@
                   (throw+ {:kaocha/early-exit 0}))
 
                 (when (:parallel config)
-                  (output/warn (str "Parallelization enabled. This is a beta "
-                                    "feature. If you encounter errors, try "
+                  (output/warn (str "Parallelization enabled. This feature is in "
+                                    "beta If you encounter errors, try "
                                     "running with the feature disabled and "
-                                    "consider filing a PR.")))
+                                    "consider filing an issue.")))
                 (when (find-ns 'matcher-combinators.core)
                   (require 'kaocha.matcher-combinators))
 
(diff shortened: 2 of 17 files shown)


alysbrooks:
A few notes:
—Somewhat unusually, this is currently rebased off of Arne's cleanup branch (#233) rather than main because I thought it might benefit from some of the cleanup work and the use of Portal.
—The basic strategy is as follows: for `kaocha.type/ns` tests (basic Clojure tests), fire off `run-tests` in a new thread using `future`, then return the futures as normal. At the top level (the test suite), recursively search for futures and deref them. That way we essentially deref at the last possible moment (after all the tests have been kicked off but before any external code that's expecting actual test results).
—One complication is that we don't want to execute in a new thread if we're already in a worker thread. I'm not sure whether nested futures are supposed to work, but it doesn't seem to work in practice and I'm not sure it would benefit us. The approach I've taken is to modify the configuration to remove `:parallel`.
—I've benchmarked this on contrived tests that favor a parallel approach (very simple tests with `Thread/sleep` calls). This is mostly to sniff out bugs (a lot of attempts early this afternoon resulted in no speedup because Kaocha was waiting for each thread to finish before running the next one); real-world tests probably won't be this parallelizable.
—I cannot run the whole Kaocha test suite with `:parallel` enabled. Ideally, you'd be able to enable this option and Kaocha just ignores it for tests that don't support it, but ensuring it can handle all the corner cases in our test suite may be too ambitious for the initial PR.

alysbrooks:
I did a little bit more testing and realized what I have doesn't work consistently. There seems to be a nondeterministic bug with loading certain specs or vars. I don't really understand it; my intuition is that all the loading would happen before we kick off any futures so this kind of bug shouldn't really happen.

We do some fancy dynamic loading (I think `jit` falls into this category, too), and it's possible some of these techniques aren't thread safe.

alysbrooks:
Another thing to look for is places we're touching atoms or dynamic vars in code that's called within the futures. Modifying a var or atom in a secondary thread isn't inherently problematic, but it's possible we've accidentally used them in a way that makes them reliant on being modified by a single thread. (For example, two atoms won't stay in sync; you should either use one atom or add some sort of other abstraction to keep them in sync.)

alysbrooks:
If you have a test suite that would benefit from parallelization, I'd be interested to hear from you. **What tests do you have?** When it's a little more polished, **would you be willing to test or benchmark parallelized test runs?**

It's easy enough to document how to turn on parallelized test run (set :parallel to true in tests.edn), but I think for this feature to be useful, the documentation will also have to include information on what kinds of tests benefit from it and what kinds of common tests aren't thread-safe. I'm hoping to hear from Kaocha users to help research this part of the documentation.

Having real-world tests will also let me determine whether parallelizing at the namespace level is enough (current approach) or if we need to consider working at the test level.

nwjsmith:
I'll be able to test and benchmark in a few days. The primary reason I'd like to see test parallelization is to speed up my suite of `test.check` tests. Parallelization at the per-test level would have the largest impact on these, but even per-namespace should provide a speed bump. Really excited to see your work on this, thank you!

alysbrooks:
@nwjsmith  You're welcome! I somehow blanked on `test.check` as a use case even though I'm a big fan of property-based testing and we use it on some of our projects, e.g., Regal. I believe `test.check` tests should work already (because they're gathered into namespace testables, which are parallelized), but enabling `test.check` tests at the test level would be another step because they're a different type.

alysbrooks:
The previous test had high variance so I quit some resource-intensive programs and set the tests to use a fixed seed:

```
↪ hyperfine 'clojure -A:test -m kaocha.runner clj' 'clojure -A:test-local -m kaocha.runner clj'
Benchmark #1: clojure -A:test -m kaocha.runner clj
  Time (mean ± σ):     43.979 s ±  0.905 s    [User: 80.840 s, System: 2.393 s]
  Range (min … max):   42.434 s … 45.172 s    10 runs
 
Benchmark #2: clojure -A:test-local -m kaocha.runner clj
  Time (mean ± σ):     39.900 s ±  1.060 s    [User: 84.105 s, System: 2.453 s]
  Range (min … max):   37.989 s … 41.194 s    10 runs
 
Summary
  'clojure -A:test-local -m kaocha.runner clj' ran
    1.10 ± 0.04 times faster than 'clojure -A:test -m kaocha.runner clj'
```

Regal isn't the ideal example for this. Even though it has a time-consuming test suite, the majority of it's time is taken up by a single test. Even though the test is a property based test that really consists of running the same simple test over a number of random inputs, from Kaocha's perspective, it's a single test.

AndreaCrotti:
I wonder if one possible alternative solution would be to support something like https://www.juxt.pro/blog/parallel-kaocha
So circleci allows you to run tests in parallel, just by running a different subset of tests in every different container.
The more tricky thing circleci does is that it merges back the test results and present the results like tests were run normally.

Maybe supporting something like this in kaocha directly would be easier and not have the same Threading performance issues?

alysbrooks:
@AndreaCrotti That's a clever solution. What changes were you thinking to make? Adding the `circleci-parallel` namespace code to Kaocha itself?

While I am interested in that solution, I plan to go ahead with the parallel option.

AndreaCrotti:
So well @alysbrooks, I guess there are three things required for this to work:
1. some way to group tests in a way that each group takes more or less the same time (which circleci does for you with from the past timing information)
2. some code to then run in parallel multiple kaochas with the various groups of namespaces
3. some code to get the results from all the kaochas and give a unified output.

Maybe only 3. could be done directly in kaocha, but probably the whole thing could be done with a plugin ideally.
For example, you could:
- run all your tests without parallelization and record the timings
- group and run all the kaochas and report everything.

One issue with the approach in this PR is that every test suite with database (or similar) writes will probably need quite a bit of adjusting to run in parallel.
This other approach I was suggesting in theory has the same issue, but if you parameterize your test resources configuration, you could also be able to just use a totally different test infrastructure for each kaocha run, and avoid the problem entirely (and it would also avoid the Threading overhead issue as well).

AndreaCrotti:
So well in short I guess it would be great to have this PR merged, but maybe I can investigate the other way to parallelize tests as a separate plugin as a possible alternative strategy.

alysbrooks:
@AndreaCrotti Interesting. I could see 2 being done with a Babashka script that we ship, and I think 3 could be a plugin, in theory. They'd be stretching the idea of plugins because the instance of Kaocha doing the combining wouldn't necessarily run any tests, it would just grab the test results and inserts them so the reporters could summarize them. Actually, a plugin that takes saved test results and feeds them back into Kaocha as though those tests just ran could be pretty interesting.

alysbrooks:
Threw together a simple graph of the Regal results:

20211_01_11 Regal Benchmarking

stig:
I'm interested in trying this in our CI. Is a snapshot published for this branch?

src/kaocha/api.clj
```diff
@@ -104,7 +104,12 @@
                                 (testable/test-seq test-plan))
                   (output/warn (str "No tests were found, make sure :test-paths and "
                                     ":ns-patterns are configured correctly in tests.edn."))
-                  (throw+ {:kaocha/early-exit 0 }))
+                  (throw+ {:kaocha/early-exit 0}))
+                (when (:parallel config)
+                  (output/warn (str "Parallelization enabled. This feature is in "
+                                    "beta If you encounter errors, try "
```
stig, line 110:
```suggestion
                                    "beta. If you encounter errors, try "
```

repl_sessions/benchmark.clj
```diff
@@ -0,0 +1,97 @@
+
+(ns benchmark
+  (:require [criterium.core :as c])
+  (:import [java.util.concurrent Executors ])
+  )
```
stig, line 5:
```suggestion
(ns benchmark
  (:require [criterium.core :as c])
  (:import [java.util.concurrent Executors ]))
```

src/kaocha/runner.clj
```diff
@@ -34,6 +35,7 @@
    [nil  "--[no-]fail-fast"     "Stop testing after the first failure."]
    [nil  "--[no-]color"         "Enable/disable ANSI color codes in output. Defaults to true."]
    [nil  "--[no-]watch"         "Watch filesystem for changes and re-run tests."]
+   [nil  "--[no-]parallel"           "Run tests in parallel. Warning: This feature is beta."]
```
stig, line 38:
```suggestion
   [nil  "--[no-]parallel"      "Run tests in parallel. Warning: This feature is beta."]
```

src/kaocha/test_suite.clj
```diff
@@ -4,10 +4,10 @@
 
 (defn run [testable test-plan]
   (t/do-report {:type :begin-test-suite})
-  (let [results (testable/run-testables (:kaocha.test-plan/tests testable) test-plan)
+  (let [results (testable/run-testables (:kaocha.test-plan/tests testable) test-plan)
```
stig, line 7: Accidentally added space at end of line.
```suggestion
  (let [results (testable/run-testables (:kaocha.test-plan/tests testable) test-plan)
```

stig (commented):
I'm excited about this PR :-)

I took the liberty to add a few stylistic nitpick suggestions.

AndreaCrotti:
> I'm interested in trying this in our CI. Is a snapshot published for this branch?

You can just point to a given sha with deps.edn and git paths?

stig:
> > I'm interested in trying this in our CI. Is a snapshot published for this branch?
> 
> You can just point to a given sha with deps.edn and git paths?

I'll see if Leiningen supports that next week, when I'm back at work. Or perhaps a hacky migration to deps.edn for the purpose of testing this is possible...? I'll have to evaluate that.

alysbrooks:
> > > I'm interested in trying this in our CI. Is a snapshot published for this branch?
> > 
> > 
> > You can just point to a given sha with deps.edn and git paths?
> 
> I'll see if Leiningen supports that next week, when I'm back at work. Or perhaps a hacky migration to deps.edn for the purpose of testing this is possible...? I'll have to evaluate that.

Leiningen doesn’t support it, as far as I remember, but the functionality is available as a plug-in.

alysbrooks:
> It seems like the use of `with-redefs` in `kaocha.plugin.capture-output/with-capture` does not fit tests running in parallel... Also, `(run! f @active-buffers)` in `kaocha.plugin.capture-output/doto-capture-buffer` leads to some confusing output when tests run in parallel.

This plugin doesn't really work well with parallelization because it was written before we introduced this feature and hasn't been updated yet. I don't think the design of `kaocha.plugin.capture-output` had threading in mind.

buhhh:
> This plugin doesn't really work well with parallelization because it was written before we introduced this feature and hasn't been updated yet. I don't think the design of kaocha.plugin.capture-output had threading in mind.

And yet the plugin is loaded by default, right? I do think the output is much worse without it though. Could Arne weigh in maybe, @plexus?

plexus:
As you can see this us still an open PR, we appreciate people testing it out and providing feedback. Most of the Kaocha codebase predates parallelization by several years, and so there will likely be more areas where we discover things don't yet play well together.

Suggestions on how to rework the plugin to work better with this branch, or perhaps a PR, are very welcome.

buhhh:
So what is the reasoning for the current implementation, (i) the use of `with-redefs` and the (ii) `active-buffers` construct?

nwjsmith:
@buhhh you can see from [the commit history that the `capture-output` plugin is based on eftest's implementation](https://github.com/lambdaisland/kaocha/commit/66f9460ae7250bb0311459e67160303a058c1da9) of output capturing. Both [`active-buffers` and the `with-redefs` are inherited from there](https://github.com/weavejester/eftest/blob/43a41b28e5981382a93503be27bce463e1b6ad4f/eftest/src/eftest/output_capture.clj).

alysbrooks:
Note to self: make sure addition of `volatile` in watch doesn't interfere with this.

alysbrooks:
I added a profiling plugin you can run by enabling `:kaocha.plugin.alpha/parallel-profiling` to `:plugins`. I don't want to add too much to this PR, but it's a plugin so the risk is relatively low.

jumarko:
Are there any plans to wrap up support for parallel test runs soon or should I resort to eftest if I need such a test runner?

AndreaCrotti:
> Are there any plans to wrap up support for parallel test runs soon or should I resort to eftest if I need such a test runner?

In case it helps you can still get some parallelism with circleci and kaocha for example https://www.juxt.pro/blog/parallel-kaocha/

src/kaocha/testable.clj
```diff
@@ -212,22 +223,68 @@
...
+(defn run-testables-parallel
+  "Run a collection of testables in parallel, returning a result collection."
+  [testables test-plan]
+  (let [load-error? (some ::load-error testables)
+        futures (doall
+                 (map (fn [t]
+                        (future
+                          (run-testable (assoc t ::thread-info (current-thread-info)) test-plan)))
+                      testables))]
+    (doall (map deref futures))))
```
marcobiscaro2112, line 265: Here we are creating an unbounded number of threads. If the test plan contains too many tests (hundreds or thousands) this will probably be a problem.

OS threads are expensive to create, and when we have a number of running threads that is significantly greater than the number of available processors, we'll also incur in a lot of context switching cost.

This probably explains what you observed in your previous comment, because for each `future` invocation, we're creating a new thread to be executed (via ExecutorService.submit):

> I noticed some cases where parallelization does indeed add overhead (I think tests ran 5 to 10 percent slower?)

> one of the methods the most time spent in was java.lang.Thread.<init>. A lot of time was spent in java.util.concurrent.AbstractExecutorService.submit ()

One alternative we could explore instead is to create a [fixed thread pool ExecutorService](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/concurrent/Executors.html#newFixedThreadPool(int)) limited to the number of [available cores](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/Runtime.html#availableProcessors()).

We then can map each test to a function that runs it, and pass that to [invokeAll](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/concurrent/ExecutorService.html#invokeAll(java.util.Collection)) (this works because a Clojure function implements Callable). This returns a list of Future, so we can do the `(doall (map deref futures))` on it the same way we're already doing.

We're still going to create a Callable and a Future per task, but we'll limit the number of threads created and the parallelism of execution to the available number of processors.

This should avoid the overhead of too much parallelism and hopefully result in parallel tests always running faster than serial tests.

What do you think?

marcobiscaro2112: Something like
```clojure
(import '(java.util.concurrent ExecutorService Executors))

(defn task [num]
  (fn []
    (.println System/out num)
    (Thread/sleep 500)
    (* num num)))

(let [num-cpus (.availableProcessors (Runtime/getRuntime))
      executor (Executors/newFixedThreadPool num-cpus)
      data (range 100)
      tasks (map task data)
      futures (.invokeAll executor tasks)
      results (doall (map deref futures))]
  (println results))
```

NoahTheDuke: I do something similar in Splint:

```clojure

(defn pmap*
  "Efficient version of pmap which avoids the overhead of lazy-seq.

  (doall (pmap (fn [_] (Thread/sleep 100)) coll)) => 3.34 secs
  (pmap* (fn [_] (Thread/sleep 100)) coll) => 202 ms"
  [f coll]
  (let [executor (Executors/newCachedThreadPool)
        futures (mapv #(.submit executor (reify Callable (call [_] (f %)))) coll)
        ret (mapv #(.get ^Future %) futures)]
    (.shutdownNow executor)
    ret))
```

alysbrooks: I couldn't remember the default executor for `future`, and it appears to be `newCachedThreadPool` so it will save some on overhead for short-lived threads, but probably creates too many in other situations.

alysbrooks: For future reference, Clojure core's fixed thread pool uses the number of processors plus 2: https://github.com/clojure/clojure/blob/08a2d9bdd013143a87e50fa82e9740e2ab4ee2c2/src/jvm/clojure/lang/Agent.java#L49C1-L51C77

buhhh:
Are there any plans to pick this up again? FWIW the current solution based on futures is good enough for our use case: we restricted the degree of parallelism by other means.

Issue #235, opened 2021-07-22
plugin-names with dots are normalized, but shouldn't


thumbnail:
According to this comment, plugins that contain a period, shouldn't get the `kaocha.plugin`-prefix.
https://github.com/lambdaisland/kaocha/blob/58d7870e65b925de5ddf029eb95b699d4fe77a78/src/kaocha/plugin.clj#L69-L75

Although in a REPL session showed:

```clj
(normalize-name :test)
=> :kaocha.plugin/test
(normalize-name :my/test)
=> :my/test
(normalize-name :my.test)
=> :kaocha.plugin/my.test
```

It's likely because the arguments need to be switched around here:
https://github.com/lambdaisland/kaocha/blob/58d7870e65b925de5ddf029eb95b699d4fe77a78/src/kaocha/plugin.clj#L69-L75

Happy to provide a patch :)

plexus:
Yes, please!

PR #236, opened 2021-07-23
Bugfix plugin name normalization

thumbnail:
fixes lambdaisland/kaocha#235.

Can be considered a breaking change; previously `:a.b.c` would be normalized to `:kaocha.plugin/a.b.c`

### DIFF
--- a/src/kaocha/plugin.clj
+++ b/src/kaocha/plugin.clj
@@ -68,7 +68,7 @@
 
 (defn normalize-name [plugin-name]
   (if (and (simple-keyword? plugin-name)
-           (not (str/includes? "." (name plugin-name))))
+           (not (str/includes? (name plugin-name) ".")))
     ;; Namespaces without a period are not valid, we treat these as
     ;; kaocha.plugin/*
     (keyword "kaocha.plugin" (name plugin-name))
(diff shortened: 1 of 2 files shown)


plexus:
Thanks! would you mind adding a CHANGELOG entry as well?

alysbrooks:
Thanks! Can you add something about this change technically being a breaking one to that CHANGELOG entry?

@plexus It occurs to me that we could add a warning if we detect that their configuration contains a plugin that will be normalized differently after this change. However, this would just be noise for people who never used a version with the bug. What do you think? (@thumbnail I wouldn't necessarily expect you to do the work of adding that warning, unless you'd like to.)

thumbnail:
I updated the CHANGELOG.md 👍 

> add a warning if we detect that their configuration [...]

This sounds like a reasonable addition. What's the possible impact for consumers for this change?

> However, this would just be noise for people who never used a version with the bug.

The warning could be shown for the first N versions (with a way to suppress it), after which the warning is removed or suppressed by default.

alysbrooks:
@thumbnail My understanding is that it previously would have tried to require `kaocha.plugin` if you specified `:a.b.c`, but now it will require just `a.b.c.`, so it's possible people had adjusted their namespaces to correct for the bug and now won't work.

I think keeping the message for just a few versions is the way to go. I don't think many people will see the message, since most plugins probably don't have dots in their name.

alysbrooks:
@thumbnail  Would you be able to make that change? If not, I can add it to your PR and merge.

thumbnail:
@alysbrooks It'd be great if you could make the additions you mentioned. Thanks :)

PR #246, opened 2021-09-13
Feature: tools-deps exec-fn support

cap10morgan:
I added a new `kaocha.runner/exec` entrypoint to allow usage from tools-dep's new -X feature. All it does is convert the `:exec-args` map to a list of CLI args and passes that to the existing `-main`.

This has the nice side effect of allowing you to load the `:test` alias in a REPL without either a) running all of your tests on load if you have a `:main-opts` directly in your alias map or b) requiring that you have a separate wrapper script that specifies the main ns to invoke to avoid a. But of course you can still have a wrapper script if you like. :)

Quick and dirty (too quick and dirty?) :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# Unreleased
+
+## Added
+
+- `kaocha.runner/exec` for use with Clojure CLI's -X feature
+
 # 1.0.887 (2021-09-01 / 38470aa)
 
 ## Added
--- a/README.md
+++ b/README.md
@@ -102,7 +102,25 @@ Add Kaocha as a dependency, preferably under an alias.
 ;; deps.edn
 {:deps { ,,, }
  :aliases
- {:test {:extra-deps {lambdaisland/kaocha {:mvn/version "1.0.887"}}}}}
+ {:test {:extra-deps {lambdaisland/kaocha {:mvn/version "1.0.887"}}
+         :exec-fn kaocha.runner/exec
+         :exec-args {}}}}
+```
+
+The `exec-args` map can contain any long-form CLI args as keyword keys.
+Use `true` as the value for args that don't take one, or `false` to get the
+`--no-whatever` equivalent.
+
+For example, this CLI invocation:
+`bin/kaocha --config-file my-tests.edn --diff-style :deep --fail-fast --no-color`
+
+...would look like this as an `exec-args` map:
+
+``` clojure
+{:config-file "my-tests.edn"
+ :diff-style  :deep
+ :fail-fast   true
+ :color       false}
 ```
 
 Add a binstub called `bin/kaocha`
@@ -110,7 +128,7 @@ Add a binstub called `bin/kaocha`
 ```
 mkdir -p bin
 echo '#!/usr/bin/env sh' > bin/kaocha
-echo 'clojure -A:test -m kaocha.runner "$@"' >> bin/kaocha
+echo 'clojure -X:test "$@"' >> bin/kaocha
 chmod +x bin/kaocha
 ```
 
--- a/README.md
+++ b/README.md
@@ -102,6 +102,26 @@ Add Kaocha as a dependency, preferably under an alias.
 ;; deps.edn
 {:deps { ,,, }
  :aliases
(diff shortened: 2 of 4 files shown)


alysbrooks:
Thanks for the PR! This looks good. However, I don't think we should drop the old `-A` approach from the README. Do you have time to add a separate section to the README?

cap10morgan:
@alysbrooks Sure. How's that?

alysbrooks:
@cap10morgan Looks good! Thank you!

200w

plexus:
Hey @alysbrooks , it seems you missed the discussion in slack. I don't think this is a good idea, and would prefer to see if we can use kaocha.repl/run for this.

We already have several different conventions for how to map command line and config flags to actual Kaocha config, this adds more ad hoc parameter wrangling, and I think with kaocha.repl/run we have what we need.

If we do want a specific exex-fn entry point then it should bypass cli-tools altogether.

alysbrooks:
@plexus yes, I did miss it, sorry about that. I can revert the merge and reopen while we discuss.

cap10morgan:
`kaocha.repl/run` might work for this (I'm not 100% sure just on a quick read of the args handling), but seems much more brittle. If it works right now, it only does so incidentally. At the very least I would want an entrypoint fn not in a `repl` namespace. This isn't a REPL and changes down the road that were beneficial for a REPL but broke this could easily happen. 

Returning a data structure summary report also isn't as ideal here since it's still just running tests at the command line.

I would also prefer to have an entrypoint fn that only took the `exec-args` map argument.

I'm happy to do this differently, but I would vastly prefer the "bypass cli-tools altogether" approach to using `repl/run`.

alysbrooks:
I don't think `kaocha.repl/run` will work with `-X` since that only takes a single map. `kaocha.repl/run-all` will, however, since it just takes the the config values (a map) to merge.

We might well change `kaocha.repl/run-all`, but it seems unlikely that we would change the first argument or remove the single-argument version.

alysbrooks:
I don't have a strong opinion on this change. My proposal would be to either:

- Not add any code at all and just document the usage of `kaocha.repl/run-all` with `-X`. (We could potentially add a Cucumber test for this specific usage.)
- Create a function in `kaocha.api`that works like `kaocha.repl/run-all`. Since that namespace is for APIs, we'd be even more hesitant to break it.

plexus:
Yes, I guess I was thinking about `kaocha.repl/run-all`. 

Let's talk about the general design of these entry points. We maintain several different ways of invoking kaocha

- kaocha.runner
- kaocha-boot
- kaocha.repl
- kaocha-nrepl (maintained by someone else)

All of these sit on top of `kaocha.api`. They implement some kind of argument/parameter handling depending on what makes sense in the context, and some kind of returns/exit value handling, depending on what makes sense in the context. `kaocha.api` is internal, and unopinionated, meant to be called by tooling or entry point implementations, not by end users.

In the end these all need to hand a test config to kaocha.api, and then they get a test result back. They should also catch `:kaocha/early-exit` and handle it, and there's some ceremony with hooks because we want plugins to be able to step in at almost any step in the process, and there are some hooks that kaocha.api can't/won't handle like post-summary or main, so it's up to the entry point to do that. In the end you end up with something like this:

``` clojure
(defn exec-fn [m]
  (try+
   (let [config (-> (config/load-config)
                    (config/merge-config (config/normalize m)))]
     (plugin/with-plugins (plugin/load-all (:kaocha/plugins config))
       (let [config (plugin/run-hook :kaocha.hooks/config config)]
         (plugin/run-hook :kaocha.hooks/main config)
         (let [result (plugin/run-hook :kaocha.hooks/post-summary (api/run config))
               totals (result/totals (:kaocha.result/tests result))
               exit-code (min (+ (:kaocha.result/error totals) (:kaocha.result/fail totals)) 255)]
           (System/exit exit-code)))))
   (catch :kaocha/early-exit {exit-code :kaocha/early-exit}
     (System/exit exit-code))))
```

```
clj -X:foo :kaocha.filter/focus '[:unit]'
```

plexus:
@alysbrooks you can stick that in `kaocha.runner`, it should do the trick.

Issue #245, opened 2021-08-26
Error counting tests in the report


Panthevm:
### Configuration

```edn
#kaocha/v1
{:tests
 [{:id            :unit
   :source-paths  ["src"]
   :test-paths    ["test"]
   :ns-patterns   ["-test$"]}]

 :fail-fast? false
 :color?     true
 :reporter   [kaocha.report.progress/report]
 :plugins    [:kaocha.plugin/filter
              :kaocha.plugin/capture-output
              :kaocha.plugin/profiling]
 :kaocha.plugin.randomize/randomize? false
 :kaocha.plugin.profiling/profiling? false}

```

### Tests

``` clojure
(ns foo.bar-test)

(clojure.test/deftest- test-my-test
  (clojure.test/is true))

(clojure.test/deftest test-my
  (test-my-test))

```

### Output

```zsh
$ make run-test
clj -A:kaocha:test
unit:   200% [=========================================================] 2/1
1 tests, 1 assertions, 0 failures.

```

does not count `clojure.test/deftest-`

200 percent of the tests were completed :)

alysbrooks:
This issue is fixed. Thanks, @humorless!

PR #248, opened 2021-09-16
fix not including private test var in building test plan - issue 245

humorless:
- fix the [issue 245](https://github.com/lambdaisland/kaocha/issues/245)

### DIFF
--- a/test/unit/kaocha/private_test.clj
+++ b/test/unit/kaocha/private_test.clj
@@ -1,0 +1,16 @@
+(ns kaocha.private-test
+  (:require [clojure.test :refer :all]
+            [kaocha.testable :as testable]))
+
+(deftest- private-test
+  (is true))
+
+(deftest public-test
+  (is true))
+
+(deftest test-load
+  (let [testable (testable/load {:kaocha.testable/type :kaocha.type/ns
+                                 :kaocha.testable/id :kaocha.private-test
+                                 :kaocha.testable/desc "kaocha.private-test"
+                                 :kaocha.ns/name 'kaocha.private-test})]
+    (is (= 3 (count (:kaocha.test-plan/tests testable))))))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+# Unreleased
+
+## Added
+
+## Fixed
+
+- Fix only considering public vars when building up the test plan
+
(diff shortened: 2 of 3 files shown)


plexus:
Looks good! Can you rebase?

alysbrooks:
This looks good, thanks @humorless! The only downside is that I believe these sample tests will be run on their own, even though they in themselves aren't testing anything.

test/unit/kaocha/private_test.clj
```diff
@@ -0,0 +1,16 @@
+(ns kaocha.private-test
+  (:require [clojure.test :refer :all]
+            [kaocha.testable :as testable]))
+
+(deftest- private-test
```
alysbrooks, line 5: I wonder if we could mark this test and the next one with the `:kaocha/pending` metadata so it doesn't get run directly in the test suite.

test/unit/kaocha/private_test.clj
```diff
@@ -0,0 +1,16 @@
...
+  (is true))
+
+(deftest public-test
```
alysbrooks, line 8: Same.

alysbrooks (approved):
Overall looks good. I left a small change. If my suggestion doesn't work, and you don't have any other ideas, feel free to merge. (Or I can.) Marking approve because I don't think my suggestion is necessary.

test/unit/kaocha/private_test.clj
humorless: When seeing `:kaocha/pending`, I thought it is smart because the two tests are not actually required. 

However, as [the document](https://cljdoc.org/d/lambdaisland/kaocha/1.0.861/doc/marking-tests-as-pending) said "Pending tests are tests that are not yet implemented, or that need fixing, and that you don't want to forget about." 

Do you think that if we use `:kaocha/pending` here, we use the semantic correctly?

plexus: I think this is fine, I actually thought it was pretty smart to do it this way. They're redundant but that shouldn't cause any issues. I wouldn't use `pending` because that will add noise to every test run. Pending is meant as a constant reminder that you have tests that need addressing, that's not really the case here.

alysbrooks: Good point. I do think including these tests in the same file is a good way to do it. I might introduce a metadata for just our test suite to skip these tests directly.

PR #249, opened 2021-09-20
Add exec fn

alysbrooks:
Adds an entry point for using Kaocha with the `exec-fn`/`-X` feature of Clojure CLI tools.

While my name's on the commit, @plexus wrote the code and @cap10morgan wrote the basis of the documentation and tests. Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,12 @@
 # Unreleased
 
 ## Added
+- `kaocha.runner/exec` for use with Clojure CLI's -X feature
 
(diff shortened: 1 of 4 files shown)


cap10morgan:
Looks like there's a discrepancy between the docs and the code: the defined fn is named `exec-fn` but the docs refer to `exec`

alysbrooks:
Thanks! I fixed it.

plexus:
Looks great @alysbrooks ! Really appreciate the doc additions as well.

PR #251, opened 2021-09-23
WIP: Memory profiling plugin

alysbrooks:
The plugin I made for looking into lambdaisland/regal#33.

Because it's not as reliable as the existing profiling plugin, we may not want to merge and maintain this. On the other thand, it is handy, and I don't think it will be a high maintenance plugin.


plexus:
This is really cool and useful and I would love to get it in!

Issue #231, opened 2021-07-09
kaocha.repl/run and run-all should catch :kaocha/early-exit


plexus:
When running through the REPL there is no handler for these exceptions, causing them to bubble up. These exceptions are internal and should never bubble out of end-user APIs.

alysbrooks:
You can reproduce this error by passing a nonexistent test ID to `kaocha.repl/run`.

alysbrooks:
Fixed by #265.

Issue #211, opened 2021-03-22
Interpret the "0 test" situation as a failure


gracechin:
Currently tests pass and no fails are thrown when there are 0 tests.

```
0 tests, 0 assertions, 0 failures.
```

Not sure if this kind of kaocha configuration is possible in the `tests.edn`, but it would be useful to have the kaocha fail when 0 tests are detected. This would help indicate that configurations are incorrect, similar to how an error is thrown, if a deftest doesn't contain any assertions. This will ensure that any mis-configurations can be brought to attention

onetom:
Recommendation on Slack, by @lread :

@onetom, I would prefer a non zero exit in this scenario as well. But... it is also pretty easy to write [a kaocha hook](https://github.com/lread/test-doc-blocks/blob/65bddd419c29f0e3ef8146aa11111df12bd7c1fd/tests.edn#L12) to achieve this.

alysbrooks:
I think this makes a lot of sense. Maybe not as default behavior, but as an option you can enable.

plexus:
This could be a good candidate for https://github.com/lambdaisland/kaocha/issues/183 , we can already introduce a flag to opt-in, then change the default for `#kaocha/v2`

frenchy64:
Along similar lines, the `No tests found` situation might also be considered a failure.

https://github.com/lambdaisland/kaocha/blob/3d834940d420d6490a2290f809526fff07ef7775/src/kaocha/api.clj#L104-L106

That might also make internal bugs like https://github.com/lambdaisland/kaocha/issues/208 more apparent.

humorless:
It seems that it can be related the PR https://github.com/lambdaisland/kaocha/pull/404

PR #265, opened 2022-01-11
Catch early-exit errors in Kaocha.repl

alysbrooks:
Addresses #231 by adding a basic message instead of letting errors bubble up.

I found calling `kaocha.repl/run` with a fake test-id was an easy way to trigger an early-exit error.

### DIFF
--- a/src/kaocha/repl.clj
+++ b/src/kaocha/repl.clj
@@ -58,12 +58,14 @@ out what's going on. The [[config]] and [[test-plan]] functions provide this
 kind of debugging information.
 
 These will particularly come in handy when developing plugins."}
-    kaocha.repl
+ kaocha.repl
   (:require [kaocha.config :as config]
             [kaocha.plugin :as plugin]
             [kaocha.api :as api]
             [kaocha.result :as result]
-            [clojure.java.io :as io]))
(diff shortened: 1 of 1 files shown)


src/kaocha/repl.clj
```diff
@@ -151,11 +153,17 @@ These will particularly come in handy when developing plugins."}
...
+     (try+
+      (->> config
+           api/run
+           (plugin/run-hook :kaocha.hooks/post-summary)
+           :kaocha.result/tests
+           result/totals)
+      (catch :kaocha/early-exit {exit-code :kaocha/early-exit}
+        (if (not= exit-code 0)
+          (output/error "Test run exited with code " exit-code)
+          (output/warn "Test run exited with code " exit-code))
```
plexus, line 165: do we ever `:early-exit` with code 0? Seems contradictory.

plexus:
Looks good! definitely an improvement!

src/kaocha/repl.clj
alysbrooks: Yeah, probably. We do if no tests are found (its own issue #211) and in the `kaocha.plugin.alpha.info` plugin.

Issue #151, opened 2020-07-08
File watching fails on MacOS Big Sur beta


rgm:
There's an issue in JNA that's biting Figwheel too for users on macOS v10.16 / v11 / Big Sur or whatever we're calling it (ie. the new version for Sept 2020 being beta tested right now). The annual increase in security-related lockdowns has introduced a crasher into JNA, which kills barbary, which kills hawk, which kills `bin/kaocha --watch`.

https://github.com/bhauman/figwheel-main/issues/253

The low-friction temp fix we figured out for Figwheel while this is being sorted out is to tell hawk to start with a `:polling` watcher instead of a `:barbary` watcher. It kills battery life since it's no longer using native filesystem change events, but at least it keeps things working while it gets sorted out at the JNA level.

Is there a similar way to allow passing hawk options into kaocha, or, if not, a PR for the same being welcome? Might be a useful add beyond this one temp fix. I think it's in Figwheel to help cope with some Docker-related issues.

alysbrooks:
@rgm Looks like this has been fixed? https://github.com/thheller/shadow-cljs/issues/767. Adding the ability to pass hawk options isn't a bad idea, so I may keep this issue open for that change.

alysbrooks:
Oh, I figured out you can already pass options into Hawk: https://cljdoc.org/d/lambdaisland/kaocha/CURRENT/doc/7-watch-mode#configuring-the-watcher

oxalorg:
Fixed in #268 and released in `v1.62.993`

PR #268, opened 2022-01-26
Add nextjournal/beholder as an alternative watcher

oxalorg:
hawk is using an old version of JNA which breaks file watching on m1 macs.
The alternative was to use :polling or :java but they were either too slow
or consumed lots of battery

[beholder](https://github.com/nextjournal/beholder) seems to be maintained, fast, and a much cleaner approach.

figwheel has also shifted to using beholder instead of hawk. shadow-cljs
also got rid of hawk

maybe beholder should just replace hawk altogether?

- [x] add beholder as default FS
- [x] hawk can be used using a configuration param `:kaocha.watch/type :hawk` but show DEPRECATED warnings
- [x] update docs and changelog

cc: #151 , https://github.com/thheller/shadow-cljs/commit/f3b89b5a3d0f35b78f0be9e457f78a70c0d7bf04 , https://github.com/bhauman/figwheel-main/pull/299

### DIFF
--- a/doc/07_watch_mode.md
+++ b/doc/07_watch_mode.md
@@ -84,7 +84,7 @@ Kaocha uses [Beholder](https://github.com/nextjournal/beholder) to watch the
 filesystem for changes. By default Beholder will pick a mechanism suitable for
 your operating system. Beholder works with OSX and M1 mac machines as well.
 
-Previously Kaocha used another watcher [Hawk](https://github.com/wkf/hawk) which
+Previously Kaocha used another watcher, [Hawk](https://github.com/wkf/hawk), which
 has been deprecated in favour of Beholder. If you still want to use it for some
 reason then use the `:kaocha.watch/type :hawk` configuration to switch to hawk.
 Please note that Hawk will be removed completely in a future release.
--- a/doc/07_watch_mode.md
+++ b/doc/07_watch_mode.md
@@ -80,17 +80,15 @@ Currently these features of `.gitignore` are not supported:
 
 ## Configuring the watcher
 
-Kaocha uses [Hawk](https://github.com/wkf/hawk) to watch the filesystem for
-changes. By default Hawk will pick a mechanism suitable for your operating
-system, but if that doesn't cut it for some reason you can switch to using
-polling instead.
+Kaocha uses [Beholder](https://github.com/nextjournal/beholder) to watch the
+filesystem for changes. By default Beholder will pick a mechanism suitable for
+your operating system. Beholder works with OSX and M1 mac machines as well.
+
+Previously Kaocha used another watcher [Hawk](https://github.com/wkf/hawk) which
+has been deprecated in favour of Beholder. If you still want to use it for some
+reason then use the `:kaocha.watch/type :hawk` configuration to switch to hawk.
+Please note that Hawk will be removed completely in a future release.
 
 Use the `:kaocha.watch/hawk-opts` configuration key to pass options to Hawk.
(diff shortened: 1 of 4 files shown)


plexus:
Yes, just replace it.

alysbrooks:
I guess the only thing we'd lose is the ability to fall back to polling, but hopefully no one is relying on that.

alysbrooks:
I realize this is a draft, but I was already looking at Beholder's source and noticed something. Does Beholder have any options? Maybe we can delete that configuration option?

src/kaocha/watch.clj
```diff
@@ -262,28 +263,50 @@ errors as test errors."
...
+  (apply beholder/watch
+         (fn [{:keys [type path]}]
+           (when (= type :modify)
+             (qput q path)))
+         (map str watch-paths)))
+
 (defn run* [config finish? q]
-  (let [hawk-opts (::hawk-opts config {})
+  (let [watcher-type (::type config :beholder)
+        watcher-opts (condp = watcher-type
```
alysbrooks, line 284: `case` seems clearer than `condp` with `=`. That's just my preference, though.

doc/07_watch_mode.md
```diff
@@ -80,17 +80,15 @@ Currently these features of `.gitignore` are not supported:
...
 
-Kaocha uses [Hawk](https://github.com/wkf/hawk) to watch the filesystem for
-changes. By default Hawk will pick a mechanism suitable for your operating
-system, but if that doesn't cut it for some reason you can switch to using
-polling instead.
+Kaocha uses [Beholder](https://github.com/nextjournal/beholder) to watch the
+filesystem for changes. By default Beholder will pick a mechanism suitable for
+your operating system. Beholder works with OSX and M1 mac machines as well.
+
+Previously Kaocha used another watcher [Hawk](https://github.com/wkf/hawk) which
```
alysbrooks, line 87:
```suggestion
Previously Kaocha used another watcher, [Hawk](https://github.com/wkf/hawk), which
```

oxalorg:
@alysbrooks Thanks for the suggestions, made those changes. Feel free to merge it in :)

oxalorg:
@alysbrooks Thanks for the suggestions, made those changes. Feel free to merge it in :)

src/kaocha/watch.clj
oxalorg: Thanks, I prefer condp 🙈

plexus: It's not matter of preference here, case is correct. It's the less powerful construct that gets the job done, and so should be preferred. Case is also faster, it can be highly optimized. Use case if you can.

oxalorg: Makes sense and yes you are right in this case (:P) case works better. But I once hit a bug something like this:

```clojure
(defn foo [x] (+ 30 x))

(let [x 5
      y 35]
  (case y
    x "not ok"
    (foo 5)"ok"
    "woops"))
```
Which seemed fine until I realised that case only works on compile time constants and does not evaluate the test cases. Since then I just defaulted to using `condp` everywhere as 90% of the time I am never checking the value directly and almost always have a predicate function.

But yeah `case` makes sense here!

oxalorg: There is more weird behavior here:
```
(let [x 5]
  (case 5
    x "ok"
    (foo 5) "wat?"
    nil))
;; => "wat?"
```
I came to the conclusion to just avoid case altogether 🙈

PR #274, opened 2022-03-08
Add :parallel-threads option to set thread pool size

john-shaffer:
The current parallel runner uses clojure.core/map with futures, which doesn't allow much control over execution. It is probably okay for unit tests, but it uses way too many threads for integration and e2e tests. We have test systems that spawn database and browser processes, and we need a much lower level of parallelism (4 seems ideal on an 8-core machine).

This uses [claypoole](https://github.com/clj-commons/claypoole)'s pmap to run the tests, which allows for simplifying the code a bit.

### DIFF
--- a/src/kaocha/runner.clj
+++ b/src/kaocha/runner.clj
@@ -186,6 +186,7 @@
   (try+
    (System/exit (apply -main* args))
    (catch :kaocha/early-exit {exit-code :kaocha/early-exit}
+     (shutdown-agents)
      (System/exit exit-code))))
 
 (defn exec-fn
--- a/src/kaocha/runner.clj
+++ b/src/kaocha/runner.clj
@@ -8,6 +8,7 @@
             [clojure.spec.alpha :as spec]
             [clojure.string :as str]
             [clojure.tools.cli :as cli]
+            [clojure.pprint :as pp]
             [expound.alpha :as expound]
             [kaocha.api :as api]
             [kaocha.config :as config]
@@ -34,6 +35,7 @@
    [nil  "--[no-]fail-fast"     "Stop testing after the first failure."]
    [nil  "--[no-]color"         "Enable/disable ANSI color codes in output. Defaults to true."]
    [nil  "--[no-]watch"         "Watch filesystem for changes and re-run tests."]
+   [nil  "--[no-]parallel"           "Run tests in parallel. Warning: This feature is beta."]
    [nil  "--reporter SYMBOL"    "Change the test reporter, can be specified multiple times."
     :parse-fn (fn [s]
                 (let [sym (symbol s)]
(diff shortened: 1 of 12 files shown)


alysbrooks:
Thank you for the PR! I've been holding off on reviewing this because I'm waiting to merge the original PR. I think being able to control the thread pool somehow is a good idea. However, I'm not sure whether bringing in another dependency is worth it.

Out of curiosity, have you run the parallel threads PR against any of your test systems?

john-shaffer:
Thanks for responding!

I considered the issue of adding a dependency, but claypoole is very light-weight and has no dependencies of its own. If that is still a problem, I could look into the possibility of vendoring just the pmap implementation. The additional control is absolutely necessary for heavy-weight test systems, both for speed and to stay within RAM constraints. It also adds a useful performance-tuning knob for others.

I have been running [this branch](https://github.com/insilica/kaocha/commits/parallelize) for testing locally and in CI for the last month. It has held up fine for hundreds of runs, although the increased parallelism did surface a [race condition](https://github.com/igrishaev/etaoin/issues/379) in etaoin.

src/kaocha/runner.clj
```diff
@@ -186,4 +186,5 @@
   (try+
    (System/exit (apply -main* args))
    (catch :kaocha/early-exit {exit-code :kaocha/early-exit}
+     (shutdown-agents)
```
plexus, line 189: Should we also shutdown agents if `-main*` exits normally?

src/kaocha/testable.clj
```diff
@@ -174,7 +189,8 @@
                  :kaocha/testable test})
             m (cond-> m
                 file (assoc :file file)
-                line (assoc :line line))]
+                line (assoc :line line)
+                thread (assoc :thread thread))]
```
plexus, line 193: We might want to namespace this? things like type/actual/expected/line/file are standard clojure.test things, if we shove extra data into reporter events we generally namespace them. Not sure how consistent we are in this but generally a good idea.

src/kaocha/testable.clj
```diff
@@ -227,6 +271,34 @@
...
+     (when others (run-testables-serial others test-plan))
+     (when nses
+       (cp/with-shutdown! [pool (cp/threadpool num-threads :name "kaocha-test-runner")]
+         (doall
+          (cp/pmap
+            pool
+            #(binding [*config*
+                       (-> *config*
+                           (dissoc :parallel)
+                           (update :levels (fn [x] (if (nil? x) 1 (inc x)))))]
```
plexus, line 292: how about `(update :levels (fnil inc 0))`

plexus:
Hi @john-shaffer , thanks a lot for this PR. There's some good stuff in here, and I appreciate you're helping to make Kaocha better. Sorry it's taken me so long before I managed to make time to have a better look.

I had a look at Claypoole, it has no dependencies of its own, it weighs itself about ~100kB. Also looking at what it does I think it's reasonable to bring that in here.

So, let's see if we can get this into a mergeable state? I see REPL code in source namespaces, a stray `tap>`, and some comments and stray whitespace that could be cleaned up. The `tap>` is actually causing CI to fail, let's see if we can get that all green as well. A CHANGELOG entry and where applicable a mention in the docs of new features (like the new command line flag) would be great.

Thanks a lot!

plexus:
Hi @john-shaffer , do you still have interest in getting this in a mergable state, or was this a drive-by PR?

john-shaffer:
Rebased on current parallelize branch, and updated a failing test case.

PR #278, opened 2022-03-17
Stop any filtering logic if a suite is already skipped

plexus:
If suite names are passed as arguments on the command line, than kaocha.config
already marks any other suite as `:kaocha.testable/skip`. This happens well
before any other filtering logic kicks in, so that any loading or other
processing can already safely be skipped.

This is the big difference between `bin/kaocha suite-name` and `bin/kaocha
--focus suite-name`. In the latter case we don't know if what you're telling us
to focus on is a suite, ns, or test var, so we have to load everything and then
process the filters. In the former part we know we can ignore any suite except
`suite-name` across the board.

This change makes the filter plugin also check for the skip flag on suites, in
which case it will do nothing for those suites. This saves a bit of work, and it
prevents unnecessary warnings. E.g. with a setup like

```
{:tests [{:id foo :focus-meta [:foo]}
         {:id :bar}]}
```

calling `bin/kaocha bar` will always emit a warning:

```
WARNING: No tests found with metadata key :foo. Ignoring --focus-meta :foo.
```

This is because the loading is skipped, so the test plan for that suite is
empty, so naturally we won't find any tests with that metadata.

### DIFF
--- a/src/kaocha/plugin/filter.clj
+++ b/src/kaocha/plugin/filter.clj
@@ -140,9 +140,11 @@
               :kaocha/tests
               (fn [suites]
                 (mapv (fn [suite]
-                        (if (or (matches? suite skip skip-meta)
-                                (and (seq focus-suites)
-                                     (not (contains? focus-suites (:kaocha.testable/id suite)))))
+                        (if (and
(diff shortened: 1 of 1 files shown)


plexus:
Don't merge yet, gonna verify this on the client project where I encountered this.

PR #279, opened 2022-03-17
Allow :kaocha.plugin.randomize/randomize? to be set at any test level

### LINKED ISSUE #272: Change randomize? at any level
## Problem 

The majority of the time, I like running my tests with `randomize? true`. Rarely, I will have a case where running tests for a specific test suite or namespace would be beneficial. Typically, this is because an earlier test(s) do an expensive, mutative operation that I do not want to redo. This might feel close to a `:once` fixture. However, there can be an important difference -- the expensive operation is the thing we actually want to test. As such, a `:once` fixture does not feel like a good fit. 

## Proposal 

I propose adding support for setting `:kaocha.plugin.randomize/randomize?` on a test suite map and namespace. 

### Suggested implementation path

[See relevant code](https://github.com/lambdaisland/kaocha/blob/ca647dbf4c04487a6186fd6654f6b618de2619e4/src/kaocha/plugin/randomize.clj#L19-L27). If the `test-plan` has `:kaocha.plugin.randomize/randomize?` set as a top-level key or in the `[:kaocha.testable/meta :kaocha.plugin.randomize/randomize?]` path, do not sort the `:kaocha.test-plan/tests`, but still continue the recursion, allowing the toggle to be on/off at any level. 

If the above sounds good, I can provide a PR.

## Alternatives

1. A harsh alternative is to set `:kaocha.plugin.randomize/randomize? false` directly in tests.edn for all tests that use that config. This is not ideal because most often you want `randomize? true`. 
2. Always pass `--no-randomize` to the CLI when running the tests. This is not ideal because it requires the caller to know they must pass that flag for tests to be idempotent. 
3. Execute the expensive operation multiple times. This is, perhaps, the cleanest path, but costs us (potentially significantly) extra time when running tests locally and on CI.

kennyjwilli:
Fixes #272 by implementing @plexus's proposal [here](https://github.com/lambdaisland/kaocha/issues/272#issuecomment-1057860081). 

Comments and questions on the implementation.
1. I added a macro `or-some` to the `randomize` namespace since it removed some repeated nestings. This macro could be moved to the `util` ns or removed altogether in favor of the nested `if-some`s. 
2. The presence of `::seed` also means that the "forced" `--no-randomize` CLI flag was not set. This could be passed explicitly by adding a new keyword. 
3. I don't quite know why `straight-sort` is called [here](https://github.com/lambdaisland/kaocha/blob/ca647dbf4c04487a6186fd6654f6b618de2619e4/src/kaocha/plugin/randomize.clj#L51), but I left it in. 
4. I tested this by manually modifying `c-tests/foo/hello_test.clj`. This was painful since REPL restarts were necessary to detect changes to that file. I assume there's a way around that, but I did not dig. Instead I wrote a unit test for `rng-sort`. I wasn't sure what the best approach for testing this was, so I left it off the draft version of the PR. Please let me know if there is a preference here. You can find the unit test below. 
```clojure
(deftest rng-sort-test
  (let [rng      ((requiring-resolve 'kaocha.plugin.randomize/rng) 123)
        rng-sort (requiring-resolve 'kaocha.plugin.randomize/rng-sort)]
    (is (= {}
           (rng-sort rng true
                     {})))
    (is (= {:kaocha.test-plan/tests [{:kaocha.plugin.randomize/sort-key -1188957731}]}
           (rng-sort rng true
                     {:kaocha.test-plan/tests [{}]})))
    (is (= {:kaocha.test-plan/tests [{}]}
           (rng-sort rng false
                     {:kaocha.test-plan/tests [{}]})))
    (is (match? {:kaocha.test-plan/tests [{}]}
                (rng-sort rng true
                          {:kaocha.testable/meta   {:kaocha.plugin.randomize/randomize? false}
                           :kaocha.test-plan/tests [{}]})))
    (is (match? {:kaocha.test-plan/tests [{}]}
                (rng-sort rng true
                          {:kaocha.plugin.randomize/randomize? false
                           :kaocha.test-plan/tests [{}]})))))
```
5. This feature should probably be documented somewhere, but I couldn't find a section about the randomize plugin, so I left that off too. Let me know how you'd like to proceed. 
6. `command-line.print-config/using---print-config` test is failing because the output got reordered due to switching map implementations. It's difficult to make the new output order match the text in `print_config.feature` since it's in a less optimal order. Suggestions?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Added
 
+- Allow :kaocha.plugin.randomize/randomize? to be set at any test level.
+
 ## Fixed
 
 ## Changed
--- a/doc/03_configuration.md
+++ b/doc/03_configuration.md
@@ -216,7 +216,7 @@ bin/kaocha --plugin profiling
 Some plugins are needed for the normal functioning of Kaocha. These are added
 automatically when using the `#kaocha/v1 {}` reader literal. They are
 
-- `:kaocha.plugin/randomize`: randomize test order
+- `:kaocha.plugin/randomize`: randomize test order ([documentation](08_plugins.md#randomize))
 - `:kaocha.plugin/filter`: allow filtering and "focusing" of tests
 - `:kaocha.plugin/capture-output`: implements output capturing during tests
 
--- a/doc/08_plugins.md
+++ b/doc/08_plugins.md
@@ -180,3 +180,54 @@ To customize which keys to print, use Kaocha's "bindings" functionality, in `tes
 ``` clojure
 :kaocha/bindings {kaocha.plugin.debug/*keys* [,,,]}
 ```
+
+## Randomize
+
+The randomize plugin picks a random seed during the `config` hook and uses that
+seed to randomize the order of test suites, namespaces, and test vars during the
+`post-load` hook. 
+
+Randomization can be toggled at the following places.
+
+1. top level config
+2. test suite config
+3. namespace metadata
+
+At each level, changing the randomize value will override the previously set 
+value. For example, you could: 
+
+- randomize by default (top level `:kaocha.plugin.randomize/randomize? true`, 
+  which is the default)
+- exclude a specific test suite by setting `:kaocha.plugin.randomize/randomize? false`
+  on the suite in tests.edn
+- but randomize a single namespace within that suite `(ns ^{:kaocha.plugin.randomize/randomize? true} ...)`
+
+Or the inverse (case 2)
+
+- turn it off at the top level
+- but turn it on for certain tests suites
+- with the exception of certain namespaces
+
+Passing the `--no-randomize` CLI flag will force all randomization to be disabled. 
+
+### Enabling
+
+The randomize plugin is enabled by default. Enable or disable randomization by
+setting the `:kaocha.plugin.randomize/randomize?` key in tests.edn at the top-level,
+in a specific test suite, or on the metadata of a namespace.
+
+### Plugin-specific command line flags
+
+```shell
+--[no-]randomize                  Run test namespaces and vars in random order.
+--seed SEED                       Provide a seed to determine the random order of tests.
+```
+
+### Plugin-specific configuration options
+
+Shown with their default values:
+
+```clojure
+#kaocha/v1
+{:kaocha.plugin.randomize/randomize? true}
+```
--- a/src/kaocha/plugin/randomize.clj
+++ b/src/kaocha/plugin/randomize.clj
@@ -16,15 +16,40 @@
                 (map straight-sort)))
     test-plan))
 
-(defn rng-sort [rng test-plan]
+(defmacro or-some
+  ([] nil)
+  ([x] x)
+  ([x & next]
+   `(let [or# ~x]
+      (if (some? or#) or# (or-some ~@next)))))
+
+(defn get-randomize
+  "Returns the randomize? value for the test plan. Metadata takes precedence
+  over test-plan."
+  [test-plan]
+  (or-some
+    (get-in test-plan [:kaocha.testable/meta ::randomize?])
+    (::randomize? test-plan)))
+
+(defn rng-sort [rng parent-randomize? test-plan]
   (if-let [tests (:kaocha.test-plan/tests test-plan)]
     (assoc test-plan
            :kaocha.test-plan/tests
-           (->> tests
-                (map #(assoc % ::sort-key (rng)))
-                (sort-by ::sort-key)
-                (map (partial rng-sort rng))))
+           (let [randomize? (or-some (get-randomize test-plan) parent-randomize?)]
+             (map (partial rng-sort rng randomize?)
+                  (if randomize?
+                    (->> tests
+                         (map #(assoc % ::sort-key (rng)))
+                         (sort-by ::sort-key))
+                    tests))))
     test-plan))
+
+(defn uses-randomize?
+  "Returns true if the test-plan uses ::randomize? at any test level."
+  [test-plan]
+  (if (get-randomize test-plan)
+    true
+    (some uses-randomize? (:kaocha.test-plan/tests test-plan))))
 
 (defplugin kaocha.plugin/randomize
   (cli-options [opts]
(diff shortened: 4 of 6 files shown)


plexus:
Hi @kennyjwilli , thanks for this! I had a first look through the diff. Not very in depth but generally I think it looks good.

> I added a macro or-some to the randomize namespace since it removed some repeated nestings. This macro could be moved to the util ns or removed altogether in favor of the nested if-somes.

That's fine, you can leave it there

> The presence of ::seed also means that the "forced" --no-randomize CLI flag was not set. This could be passed explicitly by adding a new keyword.

Can you elaborate? Does this mean that passing `--no-randomize` on the CLI does nothing if there's a fixed `::seed` in `tests.edn`?

> I don't quite know why straight-sort is called [here](https://github.com/lambdaisland/kaocha/blob/ca647dbf4c04487a6186fd6654f6b618de2619e4/src/kaocha/plugin/randomize.clj#L51), but I left it in.

Good question. Since we first store the `::sort-key` and then sort by that I also don't think this should make any difference. We might be able to take that out.

> I tested this by manually modifying c-tests/foo/hello_test.clj. This was painful since REPL restarts were necessary to detect changes to that file. I assume there's a way around that, but I did not dig. Instead I wrote a unit test for rng-sort. I wasn't sure what the best approach for testing this was, so I left it off the draft version of the PR. Please let me know if there is a preference here. You can find the unit test below.

Having a unit test like this is definitely useful, so you can go ahead and add them. I guess we could have a `.feature` test with fixed seeds to get predictable output, which would also give us a place to add docs for randomize. Not sure if you want to put that much effort in, they can be finicky to write.

You can call individual plugin hooks as functions, e.g. `(randomize-config-hook ...)`, `(randomize-post-load-hook ...)`. That's probably a good sweet spot for testing this logic, e.g.

```clj
(randomize-post-load-hook
 {:kaocha.test-plan/tests
  [{:kaocha.test-plan/id `a
    :kaocha.testable/meta {::randomize? false}
    :kaocha.test-plan/tests
    [{:kaocha.test-plan/id `a1}
     {:kaocha.test-plan/id `a2}]}
   {:kaocha.test-plan/id `b
    :kaocha.test-plan/tests
    [{:kaocha.test-plan/id `b1}
     {:kaocha.test-plan/id `b2}]}]
  ::randomize? true
  ::seed 1})
```

plexus:
> This feature should probably be documented somewhere, but I couldn't find a section about the randomize plugin, so I left that off too. Let me know how you'd like to proceed.

Seems we mainly mention the randomization feature in `04_running_kaocha_cli.md`

```
doc/04_running_kaocha_cli.md:Kaocha by default randomizes the order that tests are run: it picks a random
doc/04_running_kaocha_cli.md:Tests should be independent, but this is not always the case. This random order
doc/04_running_kaocha_cli.md:The random seed will be printed at the start of the test run. On the same code
doc/04_running_kaocha_cli.md:Use `--no-randomize` to load suites in the order they are specified, and vars in
doc/04_running_kaocha_cli.md:the order they occur in the source. You can disable randomization in `tests.edn`
```

and it's mentioned but not elaborated on in `03_configuration.md`. I think it might be good to add a dedicated section in `08_plugins`, and link to that in `03_configuration.md`, "for full details on how to use the randomization feature see the plugin docs".

> command-line.print-config/using---print-config test is failing because the output got reordered due to switching map implementations. It's difficult to make the new output order match the text in print_config.feature since it's in a less optimal order. Suggestions?

kennyjwilli:
> The presence of ::seed also means that the "forced" --no-randomize CLI flag was not set. This could be passed explicitly by adding a new keyword.
> Can you elaborate? Does this mean that passing --no-randomize on the CLI does nothing if there's a fixed ::seed in tests.edn?

Rereading this a few weeks later, I agree that the comment was not clear as all 😅  To answer your question, no, that's not what I meant. Passing `--no-randomize` on the CLI will always override any other settings. The purpose of my statement was to note the overloaded use of the `::seed` key. Its presence indicates if randomization should be used at all, and its value indicates the random seed. Those two use cases could be stored as two different keys if so desired. 

> Good question. Since we first store the ::sort-key and then sort by that I also don't think this should make any difference. We might be able to take that out.

I went ahead and removed it. https://github.com/lambdaisland/kaocha/pull/279/commits/9d5f3ed5c037c2fffabac0c9c103ddc5a4194c6e

> Having a unit test like this is definitely useful ...

Added. https://github.com/lambdaisland/kaocha/pull/279/commits/5583cda19ba6ccd7ec67a7fb675f6137e12097e4

> I think it might be good to add a dedicated section in 08_plugins

Added. LMK if you'd like any revisions, or go ahead and make any edits you'd like. https://github.com/lambdaisland/kaocha/pull/279/commits/e5f8a55baf362e8e163f84dabb99ee51fc1cbd73 and https://github.com/lambdaisland/kaocha/pull/279/commits/4edf40943136b1a087e100c08e956182fd6c2631

> Right, that's annoying... maybe do something like this? Then the order should matter less.

Ok, so I updated the test to pass (https://github.com/lambdaisland/kaocha/pull/279/commits/5cd4553160c9ebdb7c566d81a62222edaba4d443), but, for the life of me, I could not get the final block to pass, so I removed it. I pasted the failing output below, if you have any ideas on a fix. I'd also welcome advise on how to debug this. I see `features.steps.kaocha-integration`, but I'm not quite sure how that gets called, and if I can easily inject some debuggery from a repl. 

The final, failing block. 
```cucumber
    And the output should contain:
      """ clojure
       :kaocha/tests
       [{:kaocha.testable/type :kaocha.type/clojure.test,
         :kaocha.testable/id :unit,
         :kaocha/ns-patterns ["-test$"],
         :kaocha/source-paths ["src"],
         :kaocha/test-paths ["test"],
         :kaocha.filter/skip-meta [:kaocha/skip]}],
      """
```
... and the output. 
```shell
FAIL in command-line.print-config/using---print-config (test/features/command_line/print_config.feature:30)
the output should contain:
Expected: (substring? needle haystack)
Haystack:
{:kaocha/tests
 [{:kaocha.testable/type :kaocha.type/clojure.test,
   :kaocha.testable/id :unit,
   :kaocha/ns-patterns ["-test$"],
   :kaocha/source-paths ["src"],
   :kaocha/test-paths ["test"],
   :kaocha.filter/skip-meta [:kaocha/skip]}],
 :kaocha/fail-fast? false,
 :kaocha/color? false,
 :kaocha/cli-options {:config-file "tests.edn", :print-config true},
 :kaocha.plugin.randomize/seed 551409332,
 :kaocha.plugin.randomize/randomize? false,
 :kaocha/plugins
 [:kaocha.plugin/randomize
  :kaocha.plugin/filter
  :kaocha.plugin/capture-output],
 :kaocha.plugin.capture-output/capture-output? true,
 :kaocha/reporter [kaocha.report/dots]}

Needle:
 :kaocha/tests
 [{:kaocha.testable/type :kaocha.type/clojure.test,
   :kaocha.testable/id :unit,
   :kaocha/ns-patterns ["-test$"],
   :kaocha/source-paths ["src"],
   :kaocha/test-paths ["test"],
   :kaocha.filter/skip-meta [:kaocha/skip]}],
╭───── Test output ───────────────────────────────────────────────────────
│ /var/folders/t2/bmd89mm95f5cynfdtwx1l11m0000gn/T/kaocha_integration10038483014523596160 $ bin/kaocha --print-config
│ stdout:
│ {:kaocha/tests
│  [{:kaocha.testable/type :kaocha.type/clojure.test,
│    :kaocha.testable/id :unit,
│    :kaocha/ns-patterns ["-test$"],
│    :kaocha/source-paths ["src"],
│    :kaocha/test-paths ["test"],
│    :kaocha.filter/skip-meta [:kaocha/skip]}],
│  :kaocha/fail-fast? false,
│  :kaocha/color? false,
│  :kaocha/cli-options {:config-file "tests.edn", :print-config true},
│  :kaocha.plugin.randomize/seed 551409332,
│  :kaocha.plugin.randomize/randomize? false,
│  :kaocha/plugins
│  [:kaocha.plugin/randomize
│   :kaocha.plugin/filter
│   :kaocha.plugin/capture-output],
│  :kaocha.plugin.capture-output/capture-output? true,
│  :kaocha/reporter [kaocha.report/dots]}
│ 
│ stderr:
│ Downloading: lambdaisland/kaocha-cloverage/maven-metadata.xml from datomic-cloud
│ Downloading: lambdaisland/kaocha-cloverage/maven-metadata.xml from cs-mvn
│ WARNING: Implicit use of clojure.main with options is deprecated, use -M
╰─────────────────────────────────────────────────────────────────────────
1 tests, 5 assertions, 1 failures.

```

PR #282, opened 2022-04-05
Reload on :create events.

### LINKED ISSUE #277: Watch mode not working with neovim
Hi! I've been trying to use watch mode and encountered a little problem with the [beholder](https://github.com/nextjournal/beholder) library. When saving a file in vim I get (instead of the `:modify` event) two other events:

```
{:type :delete, :path #object[sun.nio.fs.UnixPath 0x20718f55 "src/main.clj"]}
{:type :create, :path #object[sun.nio.fs.UnixPath 0x352320a7 "src/main.clj"]}
```

The source code only reloads tests on `:modify`

https://github.com/lambdaisland/kaocha/blob/7ea063dcb0c43889e72e09f7f8a3f52e5773ee7a/src/kaocha/watch.clj#L275-L280

Is it possible to change it so that it also checks for `:create`?

alysbrooks:
Fixes #277 by looking for both `:create` and `:modify` events raised by beholder.

We could also look for `:delete` events. That wouldn't be necessary for the bug, but it might make more sense conceptually.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
 ## Fixed
 
 - Fix issue with `gc-profiling` plugin when there's a syntax error.
+- Ensure that modifications that are done by deleting and recreating the file
+    are picked up by using `--watch` with Beholder.
 
 ## Changed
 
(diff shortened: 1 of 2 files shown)


alysbrooks:
Regarding the CHANGELOG, I guess this is technically a change rather than a fix, but I believe it makes Beholder work like Hawk, and I think not reloading upon the creation of a brand new file is less useful.

plexus:
Since it fixes an issue that was introduced with the switch to beholder I'd say it's a regression fix.

plexus:
Can you please undo the whitespace changes?

plexus:
Thank you! Looks good to me!

PR #287, opened 2022-04-15
Update 02_installing.md

zengxinhui:
To solve this clj warning:
"WARNING: Implicit use of clojure.main with options is deprecated, use -M"


alysbrooks:
Looks good! There's another reference in the README if you want to take that. Alternatively, we can save that for another PR.

alysbrooks:
Thanks! 
thank-you-spongebob

PR #288, opened 2022-04-15
Update README.md

zengxinhui:
To solve this clj warning:
"WARNING: Implicit use of clojure.main with options is deprecated, use -M"


alysbrooks:
Thanks! 
Thanksbean

PR #290, opened 2022-04-27
WIP: Drop into REPL when tests error out

alysbrooks:
Adds a proof-of-concept error REPL when tests error out.

- [ ] Work around `capture-output` (or make `capture-output` work around plugins)?
- [ ] Try to get it to work with watch load errors?
- [ ] Add configuration for stopping on errors versus failures
- [ ] Add retry?
- [ ] Open an nREPL (or similar) for connecting to an editor.
- [ ] Include local bindings from test.

### DIFF
--- a/src/kaocha/plugin/repl.clj
+++ b/src/kaocha/plugin/repl.clj
@@ -1,0 +1,40 @@
+(ns kaocha.plugin.repl
+  (:require 
+    [clojure.main]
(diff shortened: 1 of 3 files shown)


plexus:
This has been a great proof of concept, but it doesn't look like something we want to merge as is, so I'm going to close this.

Generally the approach I'd be more interested in is creating a separate implementation of a nestable nREPL via middleware, and then having a kaocha plugin which leverages that.

Issue #289, opened 2022-04-21
Misleading warning when focussing on non-existing test


DerGuteMoritz:
When focussing on a non-existing test (e.g. due to a typo or because the test in question got renamed in the meantime), kaocha currently produces this warning:

```
WARNING: No tests were found, make sure :test-paths and :ns-patterns are configured correctly in tests.edn.
```

This is a bit misleading as it makes users think something is wrong with the test setup altogether, sending them down an unrelated rabbit hole. I assume this warning is always emitted when no tests remain after applying all filters. Maybe this could be changed so that it is only emitted when the initial set of tests (i.e. before filtering) is empty? If it's empty only after filtering, a more precise warning could be emitted instead.

Tested with 1.65.1029.

alysbrooks:
Good idea. The warning isn't meant to be describe every reason we could end up with no tests, but I think the solution you outlined makes sense.

plexus:
I think it can stay but we need to fix the behavior, if we do find tests but they are for whatever reason marked as skipped, then you should not see this message. I'd rather have it say nothing and just run 0 tests than tell the user the wrong thing. As a next step we can add specific helper messages for other scenarios, e.g. filters too strict. But we need to be careful as this thing has shown, ergonomic features like these increase complexity, thus making it more likely for things to go wrong, thus defeating their purpose.

PR #294, opened 2022-05-24
Fix "no tests" message

alysbrooks:
Previously, when filtering on a non-existent test, we'd get our existing warning about no tests being found. Unfortunately, this warning only mentions `ns-patterns` or `test-paths` (#289).

This PR fixes this by adding a new warning message for the filtering and softens the wording on the existing message.

### DIFF
--- a/src/kaocha/api.clj
+++ b/src/kaocha/api.clj
@@ -104,13 +104,12 @@
                                      (::testable/load-error %))
                                 (testable/test-seq test-plan))
                   (if (not (zero? (count (filter ::testable/skip (testable/test-seq-with-skipped test-plan)))))
-                    (output/warn (str "No tests were found and some tests were"
-                                      " skipped; check for misspelled settings"
-                                      " or combinations of settings that no tests"
-                                      " satisfy."))
-                    (output/warn (str "No tests were found. This may be an issue in your test configuration."
-                                      " To investigate, check the :test-paths and :ns-patterns keys in tests.edn."))
-                    )
+                    (output/warn (format (str "All %d tests were skipped."
+                                              " Check for misspelled settings in your Kaocha test configuration"
+                                              " or incorrect focus or skip filters.")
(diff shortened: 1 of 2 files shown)


src/kaocha/api.clj
```diff
@@ -99,12 +99,20 @@
...
             (let [config (resolve-reporter config)]
               (let [test-plan (test-plan config)]
+                
                 (when-not (some #(or (hierarchy/leaf? %)
                                      (::testable/load-error %))
                                 (testable/test-seq test-plan))
-                  (output/warn (str "No tests were found, make sure :test-paths and "
-                                    ":ns-patterns are configured correctly in tests.edn."))
+                  (if (not (zero? (count (filter ::testable/skip (testable/test-seq-with-skipped test-plan)))))
+                    (output/warn (str "No tests were found and some tests were"
```
plexus, line 107: This sounds like a contradiction, if "no tests were found" how can it be that "some of them are skipped"?

Maybe something like: "All ${count} tests were skipped. This could be due to incorrect `focus` or `skip` filters, or due to misspellings in your kaocha configuration."

src/kaocha/api.clj
```diff
@@ -99,12 +99,20 @@
...
+                                      " skipped; check for misspelled settings"
+                                      " or combinations of settings that no tests"
+                                      " satisfy."))
+                    (output/warn (str "No tests were found. This may be an issue in your test configuration."
```
plexus, line 111: Maybe better here also to talk specifically about "kaocha configuration" instead of "test configuration".

plexus (approved):
Looks good, but suggested some edits.

alysbrooks:
I took those suggestions. One potentially confusing thing is that if you select a particular test suite, it will say it skipped all the errors in the test suite, which will be less than the total number of errors in the project. I think it's still an improvement on the status quo, so merging.

PR #296, opened 2022-06-21
Add latest versions of JDK and Clojure

alysbrooks:
We're a bit behind on testing Kaocha against the latest versions of Java and Clojure. Probably no huge issues or we would have heard by now, but still worth testing, IMO.


alysbrooks:
Incidentally, Github assumes everything before the slash is a prefix, so when CircleCI concatenates the job names to produce something like `test-1.10.1-clojure/openjdk16`, it cuts everything off after the slash, so in GitHub, all we see is openjdk16.

PR #302, opened 2022-07-27
Add specs for profiling plugin data

robhanlon22:
Plugin authors (like myself) might want to consume profiling data as
part of their plugin. Having these specs available is useful for those
authors so they can write their own specs against this data without
having to violate boundaries and define the specs themselves.

### DIFF
--- a/src/kaocha/plugin/profiling.clj
+++ b/src/kaocha/plugin/profiling.clj
@@ -7,7 +7,7 @@
   (:import java.time.Instant
            java.time.temporal.ChronoUnit))
 
-(s/def ::start #(instance? Instant % ))
+(s/def ::start #(instance? Instant %))
 (s/def ::duration nat-int?)
 (s/def ::profiling? boolean?)
 (s/def ::count nat-int?)
--- a/src/kaocha/plugin/profiling.clj
+++ b/src/kaocha/plugin/profiling.clj
@@ -1,10 +1,16 @@
 (ns kaocha.plugin.profiling
   (:require [clojure.java.io :as io]
+            [clojure.spec.alpha :as s]
             [clojure.string :as str]
             [kaocha.plugin :as plugin :refer [defplugin]]
             [kaocha.testable :as testable])
   (:import java.time.Instant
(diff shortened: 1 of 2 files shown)


src/kaocha/plugin/profiling.clj
```diff
@@ -1,11 +1,17 @@
 (ns kaocha.plugin.profiling
   (:require [clojure.java.io :as io]
+            [clojure.spec.alpha :as s]
             [clojure.string :as str]
             [kaocha.plugin :as plugin :refer [defplugin]]
             [kaocha.testable :as testable])
   (:import java.time.Instant
            java.time.temporal.ChronoUnit))
 
+(s/def ::start #(instance? Instant % ))
```
robhanlon22, line 10:
```suggestion
(s/def ::start #(instance? Instant %))
```

alysbrooks:
This is a good idea. I've been thinking about the possibility of cross-plugin dependencies and how we might want to handle them if they come up. The implementation also looks solid, although I'd want to actually compare it to the plugin code and kick the tires myself before merging.

Specs add a (small) cost to future changes and they also communicate that consumers are free to rely on the current shape of the data, so I'm not sure we want to set that expectation. @plexus, what do you think?

alysbrooks:
To expand a little bit, I think my concerns could just as easily result in loosening up the specs a bit (like dropping `::profiling?`) as rejecting this PR.

robhanlon22:
or making `::profiling` into an entity as a "declaration" that the configuration value is at least present?

plexus:
No strong opinion for or against this, as long as we don't actually instrument any functions the overhead and risk should be small.

alysbrooks:
Looking a bit into commit history, it looks like the keys were all present in July 2018, soon after the plugin was initially developed. I think we're pretty unlikely to change these at this point.

I'll kick the tires a bit, but then I think we can merge it!

alysbrooks:
Thanks, @robhanlon22!

Animation of a green cartoon bean doffing its orange hat

PR #303, opened 2022-08-02
Fix exit codes

alysbrooks:
Fix some duplicated exit codes.

### DIFF
--- a/src/kaocha/config.clj
+++ b/src/kaocha/config.clj
@@ -29,7 +29,7 @@
         (update config k vary-meta assoc :replace true)
         (do
           (output/error "Test suite configuration value with key " k " should be a collection or symbol, but got '" v "' of type " (type v))
-          (throw+ {:kaocha/early-exit 252}))))
+         (throw+ {:kaocha/early-exit 250}))))
     config))
 
 (defn merge-config [c1 c2]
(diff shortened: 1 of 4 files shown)


plexus:
Nice! Thanks for picking this up. This will also need a CHANGELOG entry since it's technically a breaking change.

src/kaocha/config.clj
```diff
@@ -29,7 +29,7 @@
         (update config k vary-meta assoc :replace true)
         (do
           (output/error "Test suite configuration value with key " k " should be a collection or symbol, but got '" v "' of type " (type v))
-          (throw+ {:kaocha/early-exit 252}))))
+         (throw+ {:kaocha/early-exit 250}))))
```
plexus, line 32: Careful with whitespace!

alysbrooks:
@plexus Thanks, fixed! I was thinking about the `register` function, which has two slightly different errors (or three if you want to be even more precise) that all result in an exit code of 249. Do you think we should use separate codes for each of those? I was thinking we'd reserve the 240s (240-249 inclusive) for plugin registration errors, in case people want to catch all of them.

plexus:
I'd have to look at the code, I probably am to blame her for trying to group related or similar errors, but while debugging an issue recently it was quite frustrating to not know which of two of three cases caused the error. This is especially true if the output (eg error message) is indistinguishable.

So the question to ask is: if something goes wrong, can we pinpoint exactly which file and line caused it. If not we need to differentiate.

alysbrooks:
@plexus I separated failures to load namespaces from failures to load the plugin. The last two cases are separate paths but are only separate so we can use the correct grammar when there's one failed namespace versus multiple.

plexus:
hat_off

PR #304, opened 2022-08-09
Draft: Add summary reporter

NoahTheDuke:
Hey friends,

I'm interested in a reporter that does not print anything except for the summary at the end. I would ideally also like a leading "Starting test run" so I know it's working, but there doesn't seem to be support for that at the moment, so I don't mind excluding that.

This is pretty simple and I don't want to put a ton of work into it if there's no interest so here's the minimum working example. If there is tho, I'll write up all of the docs and changelog necessary. I couldn't find where the reporter symbols (the vectors) are tested, which is why I didn't write any tests for this PR.

Thanks so much!

### DIFF
--- a/src/kaocha/report.clj
+++ b/src/kaocha/report.clj
@@ -99,7 +99,6 @@
             [kaocha.plugin.capture-output :as capture]
             [kaocha.stacktrace :as stacktrace]
             [kaocha.testable :as testable]
-            [kaocha.testable :as testable]
             [kaocha.util :as util]
             [slingshot.slingshot :refer [throw+]]))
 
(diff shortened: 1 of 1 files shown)


plexus:
You're looking for `kaocha.report/result`

NoahTheDuke:
> You're looking for `kaocha.report/result`

I don't see it referenced anywhere in the docs. Would you accept a PR mentioning it?

plexus:
Yeah sure, that would be great. It's mostly used as part of other reporters, I guess I never imagined people using it by itself but it makes sense.

alysbrooks:
@NoahTheDuke I think I'll close this PR so you can open a separate one with the do change. Or are you planning to repurpose this PR?

NoahTheDuke:
Sorry, been busy at work. I'll open a new PR with the documentation.

PR #309, opened 2022-09-05
Improve warning when notification can't appear.

alysbrooks:
Add some helpful context when the system isn't able to display a notification message. From my observations on CircleCI, it appears the HeadlessException doesn't actually fire. Not sure why!

I realized after making and committing this change that an alternative/additional approach would be to disable notifier when the CI environment variable is set. I wasn't able to think of any situations where CI would be set and the user would want to see a notification, but it probably merits more thought. I want to be very careful before applying this kind of "smart" configuration because if it fails, it can be confusing and more frustrating than an extraneous warning.

### DIFF
--- a/doc/plugins/notifier_plugin.md
+++ b/doc/plugins/notifier_plugin.md
@@ -9,10 +9,10 @@ useful in combination with `--watch`, e.g. `bin/kaocha --plugin notifier
 --watch`.
 
 It does this by invoking a shell command which can be configured, so it can be
-used to invoke an arbitrary command or script. By default it will try to
+used to invoke an arbitrary command or script. By default, it will try to
 detect which command to use, using either `notify-send` (Linux) or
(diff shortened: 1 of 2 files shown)


plexus:
I'm tempted to go the other way around, and remove the AWT error from the warning. It doesn't give the user any more actionable information. And that way the warning gets less noisy and people are more likely to just ignore it on CI, which is what we want. Currently we're sending mixed signals because in the same line we're saying "warning" and "error". It's not an error, it's a condition we detect and handle appropriately, no further action needed.

alysbrooks:
Dropped the exceptions. I think avoiding warnings in CI is a discernible difference because it makes the logs less noisy. At the same time, making the warning longer also makes the warning take up more space in the log, so I think removing that part makes sense. It probably would make sense to add something to the docs instead because if someone wants to eliminate as many warnings as possible, they'd probably look through the docs or search for the error text.

PR #314, opened 2022-09-21
Ensure --watch respects the profile on reload.

### LINKED ISSUE #310: Watch mode ignores :skip-meta
When watch mode is started with a profile that has `:skip-meta` set in tests.edn, pressing `Enter` runs all the tests, without skipping anything.

alysbrooks:
When using profiles with --watch, the initial configuration would be loaded with the desired profile, but subsequent ones would use the :default profile. This is because it would use the initially parsed configuration to run the tests, but then, upon rerunning the test, reload the configuration without taking into account the profile.

Fixes #310

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 ## Fixed
 
+- Fix configuration parsing when using `--watch`. Previously, profiles would be
+    respected on the initial load, but not after watch reloaded the
+    configuration.
+
 ## Changed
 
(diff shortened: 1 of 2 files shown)


alysbrooks:
I considered writing a test for this, but it would be fairly complex. I could write tests for watch.reload-config, although I think we might want to refactor it, actually.

plexus:
We should disable the codecov checks if we routinely ignore them.

PR #315, opened 2022-09-21
Refactor configuration loading to reduce duplicated code

alysbrooks:
There's duplicated config logic across `kaocha.runner` and `kaocha.watch`.

In addition to simplifying maintenance, this change would:
* Avoid future issues where configuration loading diverges in `kaocha.runner` and `kaocha.watch`, like #310.
* Provide an easy way to work with configuration at the REPL.
* Allow watch to display errors when an invalid configuration is reloaded.

Relatedly, we may be able to address this TODO:

```
   ;; TODO: we're calling the config hook here in multiple places, and it's also
  ;; being called in `kaocha.api`. Given that we already need the fully expanded
  ;; config here (since now we're potentially adding suites in the config hook),
  ;; we should call it once at the top here, and pass the processed config into
  ;; kaocha.api. Punting on that because it requires a coordinated update in
  ;; kaocha.repl and kaocha-boot.
```

I suspect when `kaocha.watch` was first created it was easy enough to keep these two places in sync and most of the complexity has been added since.

### DIFF
--- a/src/kaocha/config.clj
+++ b/src/kaocha/config.clj
@@ -228,8 +228,6 @@
   and displaying messages about any errors."
   ([config-file]
    (load-config2 config-file nil {} nil nil))
-  ([config-file profile]
-   (load-config2 config-file profile {} nil nil))
   ([config-file profile opts]
    (load-config2 config-file profile opts nil nil))
   ([config-file profile opts cli-options cli-args]
--- a/src/kaocha/config.clj
+++ b/src/kaocha/config.clj
@@ -256,11 +256,6 @@
 (defn plugin-chain-from-config [config cli-options]
   (plugin/load-all (concat (:kaocha/plugins config) (when cli-options (:plugin cli-options)))))
 
-(defn reload-config [config plugin-chain]
-  (if-let [config-file (get-in config [:kaocha/cli-options :config-file])]
-    (let [profile (get-in config [:kaocha/cli-options :profile])]
-      [(load-config2 config-file profile {}) plugin-chain])
-    [config plugin-chain]))
 
 (defn resolve-reporter [reporter]
   (cond
--- a/src/kaocha/config.clj
+++ b/src/kaocha/config.clj
@@ -224,7 +224,8 @@
     config))
 
 (defn load-config2
-  "Loads config, factoring in profiles, and handling errors."
+  "Loads config from config-file, factoring in profile specified using profile,
+  and displaying messages about any errors."
   ([config-file profile]
    (load-config2 config-file profile {} nil nil))
   ([config-file profile opts]
--- a/src/kaocha/config.clj
+++ b/src/kaocha/config.clj
@@ -226,6 +226,8 @@
 (defn load-config2
   "Loads config from config-file, factoring in profile specified using profile,
   and displaying messages about any errors."
+  ([config-file]
+   (load-config2 config-file nil {} nil nil))
   ([config-file profile]
    (load-config2 config-file profile {} nil nil))
   ([config-file profile opts]
--- a/src/kaocha/config.clj
+++ b/src/kaocha/config.clj
@@ -149,7 +149,6 @@
                    #(merge {:resolver aero/resource-resolver} %))
            (read-config-source resource)))))
 
-
 (defn load-config-file
   "Loads and returns configuration from `source` or the file \"tests.edn\"
   if called without arguments.
@@ -247,7 +246,7 @@
                   (catch AssertionError e
                     (output/error "Invalid configuration file:\n"
                                   (.getMessage e))
-                    (throw+ {:kaocha/early-exit 252}))) ]
+                    (throw+ {:kaocha/early-exit 252})))]
      (cond-> config
        check_config_file (update ::warnings conj check_config_file)
        check (update ::warnings conj check)))))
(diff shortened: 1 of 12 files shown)


alysbrooks:
We can probably handle #319 in this same PR.

src/kaocha/config.clj
```diff
@@ -217,6 +222,39 @@
                           tests)))))
     config))
 
+(defn load-config2
```
plexus, line 225: Let's pick something descriptive, this does quite a bit now so we have several options:

- `load-config-with-checks`
- `load-cli-config`
- `load-config-for-profile`
- `load-config-and-validate`

Looking at it this way I think we can factor this much better as a multi-stage thing

I think we can keep profile in `opts`, no need to introduce an extra arity for that. But maybe make `read-config-source` deal more gracefully with `{:profile nil}`, using the same fallback as when the profile is absent. That would also simplify the calling code. (you can then do `{:profile profile}` instead of `(if profile {:profile profile} {})`

```clj
(defn apply-cli [config cli-opts cli-args]
  (-> config
      (apply-cli-opts cli-opts)
      (apply-cli-args cli-args)))

(defn validate! [config config-file]
  ;; print warning
  ;; check spec, might throw
  )

(defn load-config-for-cli-and-validate [config-file opts]
  (-> config-file
      (load-config opts)
      (apply-cli cli-opts cli-args)
      (validate! config-file)))
```

src/kaocha/config.clj
```diff
@@ -217,6 +222,39 @@
...
+   (load-config2 config-file nil {} nil nil))
+  ([config-file profile opts]
+   (load-config2 config-file profile opts nil nil))
+  ([config-file profile opts cli-options cli-args]
+   (let [config (cond-> config-file
+                    true (load-config-file (if profile (assoc opts :profile profile) opts))
+                    cli-options (apply-cli-opts cli-options)
+                    cli-args (apply-cli-args cli-args))
+
+         check_config_file (when (not (. (io/file (or config-file "tests.edn")) exists))
```
plexus, line 238: Dashes and `(.exists ...)` would be more idiomatic here. The `(. ... foo)` syntax is really only used in Clojure core code while bootstrapping.

src/kaocha/config.clj
```diff
@@ -217,6 +222,39 @@
...
+                             (output/warn (format (str "Did not load a configuration file and using the defaults.\n"
+                                          "This is fine for experimenting, but for long-term use, we recommend creating a configuration file to avoid changes in behavior between releases.\n"
+                                          "To create a configuration file using the current defaults and configuration file location, create a file named %s that contains '#%s {}'.")
+                                                  config-file
+                                     current-reader)))
```
plexus, line 243: Please pay attention to whitespace. Your editor should be able to re-indent this.

src/kaocha/config.clj
```diff
@@ -217,6 +222,39 @@
...
+         check  (try
+                  (specs/assert-spec :kaocha/config config)
+                  (catch AssertionError e
+                    (output/error "Invalid configuration file:\n"
+                                  (.getMessage e))
+                    (throw+ {:kaocha/early-exit 252})))]
+     (cond-> config
+       check_config_file (update ::warnings conj check_config_file)
+       check (update ::warnings conj check)))))
```
plexus, line 252: Not sure I understand what this is meant to the doing. Both `check_config_file` and `check` will AFAICT return `nil` or throw an exception.

src/kaocha/runner.clj
```diff
@@ -148,29 +148,12 @@
 
   (binding [spec/*explain-out* expound/printer]
     (let [{{:keys [config-file plugin profile]} :options} (cli/parse-opts args cli-options)
-          config                                          (-> config-file
-                                                              (or "tests.edn")
-                                                              (config/load-config (if profile
-                                                                                    {:profile profile}
-                                                                                    {})))
-          _check_config_file                              (when (not (. (File. (or config-file "tests.edn")) exists))
```
plexus, line 156: I see now where the underscores come from. And I guess the `.` syntax was already here. These sections are really only here for side effects I believe, hence why the names start with an `_`.

src/kaocha/repl.clj
```diff
@@ -81,11 +81,7 @@ These will particularly come in handy when developing plugins."}
    (config {}))
   ([extra-config]
    (let [config-file (:config-file extra-config "tests.edn")
-         config      (-> (config/load-config config-file)
-                         (config/merge-config (config/normalize extra-config))
-                         (cond-> #_config
-                           (.exists (io/file config-file))
-                           (assoc-in [:kaocha/cli-options :config-file] config-file)))
+         config      (config/load-config2 config-file)
```
plexus, line 84: It seems this loses the `extra-config`? this is supposed to get merged into the config, so you can locally override bits of it.

plexus (changes requested):

alysbrooks:
Made those changes. I think we're converging on the right solution, but Joshua said he'd have other suggestions. Hopefully, that'll be the last round of changes. :)

plexus:
Seems like this is a big improvement, let's get this shipped. For the remaining inconsistencies let's open a new issue for discussion.

@alysbrooks it seems you have a conflict, can you rebase?

plexus:

```clj
```

PR #318, opened 2022-10-11
Add Java 17 and 19 to test matrix

alysbrooks:
Adds the latest LTS (17) and the latest interim release (19).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,8 +1,8 @@
 version: 2.1
 
 orbs:
-  kaocha: lambdaisland/kaocha@0.0.1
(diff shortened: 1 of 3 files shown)


alysbrooks:
I don't know whether this is a good long-term strategy. I will merge this (to get tests working on a recent JVM) and then we can revisit.

alysbrooks:
Actually, this might be the only test that has EDN output, so it might just be a matter of documenting the current behavior. We might add others down the road (I've thought about adding EDN output to the profiling plugins), but I don't think it will ever be a major architectural or interface decision.

PR #325, opened 2022-10-20
fix unreachable case when :kaocha.spec.test.check/syms is a collection

colinkahn:
`condp` using `cond`-like `:else`, resulting in a "no case found" error when `:kaocha.spec.test.check/syms` is a collection.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
   respected on the initial load, but not after watch reloaded the configuration.
 - Notifier now reports errors in your notification command instead of silently
   failing.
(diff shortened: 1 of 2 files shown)


alysbrooks:
Good catch, @colinkahn! Ideally the changelog entry would be a little bit clearer about the issue itself—perhaps mentioning the error message? As currently written, I'm not sure someone running into the bug would recognize it in the description.

colinkahn:
@alysbrooks does this sound any clearer:

- Fixed `Failed loading tests: Exception: java.lang.IllegalArgumentException: No matching clause: []`  exception when `:kaocha.spec.test.check/syms` is a collection

PR #327, opened 2022-10-26
Correct documentation of `wrap-run` hooks

leo-ppp:
This documentation is misleading; `wrap-run` hooks do not receive two arguments but a single one, being `run`. We can see this here

https://github.com/lambdaisland/kaocha/blob/c3b4bfb78287988c1bdba8827c0e4369252ca8fd/src/kaocha/plugin/hooks.clj#L90-L91

Writing a wrap-run hook that takes two arguments will throw an `ArityException` when running tests.


alysbrooks:
Thanks @leo-ppp!

alysbrooks:
GIF of someone writing THANK YOU out on a piece of paper and adding little triangular flairs around the edge

PR #334, opened 2022-11-19
Update stability statement in docs

### LINKED ISSUE #333: Docs Introduction implies that kaocha is still in alpha status.
From https://cljdoc.org/d/lambdaisland/kaocha/1.71.1119/doc/1-introduction
> Currently Kaocha's versioning scheme is `0.0-${commit count}`, and releases are made often. As long as the version is at `0.0` Kaocha will be considered alpha, in other words: subject to change.

Seeing as the current version is 1.71.1119, this seems to be deceptively underselling the maturity and expected stability of the project for potential new users.

alysbrooks:
Fixes #333. I think this change is pretty easy, but I could see some debate about how to describe our attitude toward breaking changes.

### DIFF
--- a/doc/01_introduction.md
+++ b/doc/01_introduction.md
@@ -37,9 +37,9 @@ Features include:
 - Extensible through plugins
 - Tool agnostic (Clojure CLI, Leiningen, boot)
 
-Currently Kaocha's versioning scheme is `0.0-${commit count}`, and releases are
-made often. As long as the version is at `0.0` Kaocha will be considered alpha,
-in other words: subject to change. Keep an eye on the CHANGELOG.
+Currently Kaocha's versioning scheme is `1.${release count}-${commit count}`, and releases are
+made often. Kaocha is stable, but we occasionally release breaking changes. We try to minimize 
+the impact and the number of people affected, but it's still good to keep an eye on the CHANGELOG.
 
 Kaocha requires Clojure 1.9. ClojureScript support requires Clojure and
 ClojureScript 1.10.
--- a/doc/01_introduction.md
+++ b/doc/01_introduction.md
@@ -37,9 +37,12 @@ Features include:
 - Extensible through plugins
 - Tool agnostic (Clojure CLI, Leiningen, boot)
 
-Currently Kaocha's versioning scheme is `1.${release count}-${commit count}`, and releases are
-made often. Kaocha is stable, but we occasionally release breaking changes. We try to minimize 
-the impact and the number of people affected, but it's still good to keep an eye on the CHANGELOG.
+Currently Kaocha's versioning scheme is `1.${release count}-${commit count}`,
+and releases are made often. Kaocha is stable and we try to avoid breaking
+changes. If they aren't avoidable, we minimize their impact and the number of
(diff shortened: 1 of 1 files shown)


plexus:
Yeah that makes sense, I think we should at least mention that we "generally try our best to avoid breaking changes".

I think with Kaocha we're a bit more relaxed about breaking changes because it's an end-user tool, and you sometimes want to evolve how it works for better convience and UX, whereas with libraries we have a much harder commitment not to break existing code that depends on it. With libraries we can also always branch out into new namespaces or even new artifact names so both versions can co-exist, we wouldn't really want to fork Kaocha in that way.

That said, Kaocha is *also* a library in a sense, in that it's intended as much for other tool authors or plugin authors to build on as it is an end user tool. And so for those interfaces we do have a fairly strong commitment not to break third party code that leverages Kaocha.

alysbrooks:
@plexus Does this sound better? I wonder if we should create a Lambda Island-wide policy for breaking changes.

plexus:
Sure, having that documented somewhere in `lambdaisland/open-source`, together with documentation about the versioning scheme we use would be good.

The gist of it is
- we take backwards compatibility very seriously (don't break userspace!), and avoid breaking changes if at all possible
- projects that are still in version `0.x` are exempted, here we retain the right to "get it right" before locking it down
- in practice though the main thing that matters is amount of adoption and thus potential breakage. If a `0.x` project is already in widespread use then it's unlikely we'll make breaking changes, and we should bump it to `1.x` at the next release.
- there's a tiny bit of wiggle room for parts of the API that we don't consider public, or for inputs that are not (yet) officially supported or specified, similar to what Clojure does. (e.g. we may have a function accept a type of argument that in the past would cause an exception.) It still hinges on impact assessment. We should expect this to not impact any users, unless they are really doing things they should definitely not be doing.
- if we do see a legitimate need for breaking changes, then we consider leaving the original namespaces and artifact id "frozen", and instead fork to a new major version with its own namespace names and artifact id, so that at least in principle both can live side by side in the same project without impacting each other. We've done this so far for `deep-diff2` (changes were necessary for clojurescript compat), and kaocha-cljs2 (major rewrite with different underpinnings).
- projects with an end user UI (including terminal UIs like kaocha) are a special case, we retain the right to improve the UI and general end user experience, even if this means changing existing behavior. In these cases we generally expect anyone who is building tooling on top of these projects to use underlying APIs, which are generally provided (e.g. kaocha.api) rather than trying to interface/script the UI.

Versioning (this has already been described somewhere, but I'm not sure where :see_no_evil: )

- `major.releases.commits`
- `major` is 0 for work-in-progress/alpha, 1 for stable. It generally stays at one, unless we fork the project into a new artifact (e.g. deep-diff2 has major version 2)
- releases = number of releases to dates
- commits = total number of commits

alysbrooks:
@plexus I can look into adding that. In the meantime, does the new wording look okay.

plexus:
Looks good! I'm adding some of my comments to the lambdaisland/open-source README, so in the future we can link there.

plexus:

```clj
```

PR #369, opened 2022-12-18
Fix `Capability check for org.clojure/tools.cli` doc formatting

frenchy64:
The title and first sentence of this doc page are incorrectly combined: https://cljdoc.org/d/lambdaisland/kaocha/1.71.1119/doc/capability-check-for-org-clojure-tools-cli-if-a-project-s-dependency


alysbrooks:
Thanks! Do you mind reformatting the paragraph to be wrapped to 80 characters? If not, I can do that.

PR #371, opened 2022-12-19
Ensure unskipped testable for reporting load error

### LINKED ISSUE #370: Kaocha watch mode not reporting load errors
When there are 2 (or more) entries in `:kaocha/tests` and you activate the second one in watch mode, if reloading a file fails, the stacktrace is not printed.

Please find a full reproduction here: https://github.com/frenchy64/reprod-kaocha-watch-bug

frenchy64:
Fix #370 

If the first testable is already skipped in watch mode, then the load error is never reported. Tested successfully in the reproduction mentioned in #370.


alysbrooks:
I ran through your example with the fix applied, and it worked. Thanks for providing that—it helps a lot! I also verified that, once the issue preventing reloads is fixed, the correct tests are still skipped. (If you're interested in the changes I made to test that scenario, I can submit a PR.)

This fix makes sense, although I wonder if a better algorithm would be to arbitrarily pick the first *unskipped* testable and put the load error on that. I think doing it that way is technically more correct (since we aren't activating a previously skipped testable), but I'm not sure there would be any user-facing impact. In theory, some plugin or hook could depend on whether tests are skipped, even in the event of a load error, so it seems preferable.

If you want to try that, @frenchy64, go ahead. Otherwise, I think I'll merge this change for now since I think it's an improvement.

frenchy64:
@alysbrooks sure, I went for the minimal fix. I can implement your suggestion.

frenchy64:
@alysbrooks I implemented you suggestion and checked that it still works. Now the message is improved to include the suite I selected:

```
ERROR in checker (loads_once.clj:3)
Failed reloading loads-once:
```

src/kaocha/watch.clj
```diff
@@ -244,15 +244,20 @@ errors as test errors."
...
+                          (into [] (map (fn [suite]
+                                          (if (and (not @applied?)
+                                                   (not (::testable/skip suite)))
+                                            (do (vreset! applied? true)
+                                                (assoc suite
+                                                       ::testable/load-error error
+                                                       ::testable/load-error-file (or file (util/ns-file error-ns))
+                                                       ::testable/load-error-line line
+                                                       ::testable/load-error-message (str "Failed reloading " error-ns ":")))
+                                            (assoc suite ::testable/skip true))))
```
alysbrooks, line 259: Just to make sure I understand, this line is required to retain the "skip the rest" behavior?

alysbrooks (approved):
Looks good. I don't see volatiles used much (actually, I can't think of a use I've seen in actual Clojure code), but they seem appropriate here. I just have one question I'd like to get an answer to before merging.

alysbrooks:
Actually, another thing. There are no tests for this change. Would you be interesting in writing some? It's an awkward change to write tests for, so I'm okay merging this without them. If you are interested, you might be able to model them after `kaocha.watch-test`. Thanks!

frenchy64:
Yep sure I'll write some tests.

src/kaocha/watch.clj
frenchy64: Yes exactly.

frenchy64:
@alysbrooks ok done. I made a little framework that should make these kinds of tests easier.

frenchy64:
Scratch that, the test also passes on the master branch. Need to revise.

frenchy64:
Fixed it. I forgot it the reproduction needs to first have a passing test run before introducing the compilation error. The test now fails on master but passes on this branch.

test/shared/kaocha/integration_helpers.clj
```diff
@@ -126,6 +129,123 @@
...
+  *in* is bound to the output stream of this Process
+  - eg., use (read-line) to read the output of ./bin/kaocha
+  *out* is bound to the input stream of this Process
+  - eg., use (println) to send input to ./bin/kaocha
+  
+  Tips:
+  - avoid using `testing` or `is` in ways that can swallow exceptions
+  in body. Try and make the first exception that's thrown contain
+  all the debugging information you need using the helpers in the
+  rest of this file."
```
alysbrooks, line 192: This is a really helpful docstring, thanks!

test/shared/kaocha/integration_helpers.clj
```diff
@@ -126,6 +129,123 @@
...
+  `(interactive-process* ~m ~args #(do ~@body)))
+
+(comment
+  (zero?
+    (interactive-process {:dir "." :runner (io/file "echo")} ["hello"]
+                         (let [l1 (read-line)
+                               _ (assert (= "hello" l1) (pr-str l1))])))
+  )
+
+(defn read-string-line
```
alysbrooks, line 203: Maybe `read-line-or-throw`?

test/shared/kaocha/integration_helpers.clj
```diff
@@ -126,6 +129,123 @@
...
+
+  If not, slurps the rest of the output for debugging purposes and throws an exception."
+  [lines]
+  (mapv (fn [l]
+          (let [s (read-string-line)]
+            (or (is (= l s))
+                (throw (ex-info (format "Failed to match %s\nEntire expected: %s\nRest of stream:\n%s"
+                                        (pr-str l) lines (str/split-lines (slurp *in*)))
+                                {})))))
+        lines))
```
alysbrooks, line 220: Could `match?` (from matchers/combinators) be used instead of this function? `kaocha.watch-test` already uses this in `reload-config-test`.

test/shared/kaocha/integration_helpers.clj
```diff
@@ -126,6 +129,123 @@
...
+  eg., [\"--watch\" \"--focus\" \"foo\"]
+  
+  Kills the process after 30 seconds if it has not already
+  been exited. Ensures the process is killed immediately after
+  body is executed. Returns the exit code.
+  
+  Executes body in the following environment:
+  *process* is the running Process executing `./bin/kaocha ~@args`
+  *in* is bound to the output stream of this Process
+  - eg., use (read-line) to read the output of ./bin/kaocha
```
alysbrooks, line 184:
```suggestion
  - e.g., use (read-line) to read the output of ./bin/kaocha
```

test/shared/kaocha/integration_helpers.clj
```diff
@@ -126,6 +129,123 @@
...
+                (Thread/sleep timeout-ms)
+                @kill)
+              (binding [*in* r
+                        *out* w]
+                (f)))))
+        (assert (.waitFor p 10 (java.util.concurrent.TimeUnit/SECONDS))
+                "Process failed to stop!")
+        (.exitValue p)
+        (finally
+          (is (not (realized? kill)) (format "Process was killed after %sms timeout!!" timeout-ms))
```
alysbrooks, line 163:
```suggestion
          (is (not (realized? kill)) (format "Process was killed after %sms timeout!" timeout-ms))
```

We usually don't use exclamation points, but if you are going to, use just one. I'm marking all of them if you want to use GitHub's feature to pull in reviewer comments into a commit.

test/shared/kaocha/integration_helpers.clj
```diff
@@ -126,6 +129,123 @@
...
+          (try (.exitValue p)
+               (catch IllegalThreadStateException _
+                 (is nil "Process was killed after executing entire test suite!!")
```
alysbrooks, line 166:
```suggestion
                 (is nil "Process was killed after executing entire test suite!")
```

test/shared/kaocha/integration_helpers.clj
```diff
@@ -126,6 +129,123 @@
...
+    (interactive-process {:dir "." :runner (io/file "echo")} ["hello"]
+                         (let [l1 (read-line)
+                               _ (assert (= "hello" l1) (pr-str l1))])))
+  )
+
+(defn read-string-line
+  "Read a line from the current integration test and throw if the process has died."
+  []
+  (or (read-line)
+      (throw (ex-info "Process killed!!" {}))))
```
alysbrooks, line 207:
```suggestion
      (throw (ex-info "Process killed!" {}))))
```

test/shared/kaocha/integration_helpers.clj
```diff
@@ -126,6 +129,123 @@
...
+  [f]
+  ;; can't recur across try, use atom
+  (let [seen (atom [])]
+    (try
+      (loop []
+        (let [s (read-string-line)]
+          (when-not (f s)
+            (swap! seen conj s)
+            (recur))))
+      (catch Throwable e
```
alysbrooks, line 249: If you're just just expecting `ex-info` `Throwable`s, maybe put `clojure.lang.ExceptionInfo`? We also use slingshot, but I can't think of an idiomatic way to make this catch more specific.

test/shared/kaocha/integration_helpers.clj
```diff
@@ -126,6 +129,123 @@
...
+              .start)
+        kill (delay (.destroy p))
+        timeout-ms (cond-> 30000
+                     (System/getenv "CI") (* 2))]
+    (binding [*process* p]
+      (try
+        (let [input (.getOutputStream p)
+              output (.getInputStream p)]
+          (with-open [w (io/writer input)]
+            (with-open [r (io/reader output)]
```
alysbrooks, line 151: I think you can combine these into a single `with-open`.

test/unit/kaocha/watch_test.clj
```diff
@@ -186,3 +187,57 @@
...
+                    "[watch] Error reloading, all tests skipped."])
+                 ;; fix the compilation error...
+                 (spit-good-test)
+                 (dbg "after good-test")
+                 (integration/expect-lines
+                   ["[watch] Reloading #{bar-test}"
+                    "[(.)]"
+                    "1 tests, 1 assertions, 0 failures."])
+                 (finally
+                   ;; unsure how to exit process via (println) ... eg., how to enter ^C ?
```
alysbrooks, line 241: What you have seems to work, but maybe if *process* were a map in an atom, you could have a `:process` key containing an actual `Process` and `:continue` key telling it whether to keep going? I think this might make `*interactive-process*` a little tidier, too.

We can also save this reworking for a later PR since I don't want to hold up this bug fix too much longer.

alysbrooks (commented):
Thank you for all your work on the tests! I think we might be able to use your helpers to test some other behaviors of `--watch`.  I think after these changes, this'll be ready to merge!

test/shared/kaocha/integration_helpers.clj
```diff
@@ -126,6 +129,123 @@
...
+(def ^:dynamic ^Process *process* nil)
+
+(defn interactive-process*
+  [{:keys [dir runner] :as _m} args f]
+  (let [p (-> (doto (ProcessBuilder. ^java.util.List (cons (str runner) args))
+                (.directory (io/file dir)))
+              .start)
+        kill (delay (.destroy p))
+        timeout-ms (cond-> 30000
+                     (System/getenv "CI") (* 2))]
```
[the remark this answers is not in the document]
frenchy64: It didn't seem to help, no. I really don't know, this was just a hunch.

test/shared/kaocha/integration_helpers.clj
frenchy64: I've never used it, but I changed it for you.

test/unit/kaocha/watch_test.clj
frenchy64: I pasted your suggestion in this comment for later.

test/shared/kaocha/integration_helpers.clj
```diff
@@ -126,6 +129,123 @@
...
+  Executes body in the following environment:
+  *process* is the running Process executing `./bin/kaocha ~@args`
+  *in* is bound to the output stream of this Process
+  - eg., use (read-line) to read the output of ./bin/kaocha
+  *out* is bound to the input stream of this Process
+  - eg., use (println) to send input to ./bin/kaocha
+  
+  Tips:
+  - avoid using `testing` or `is` in ways that can swallow exceptions
+  in body. Try and make the first exception that's thrown contain
```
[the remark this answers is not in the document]
frenchy64: Reworded.

frenchy64:
@alysbrooks thanks for the review, I think I've addressed everything.

frenchy64:
Hmm looks like the test can exceed CircleCI's 10 minute build limit: https://app.circleci.com/pipelines/github/lambdaisland/kaocha/1030/workflows/75013baa-25a5-4f2a-9aa9-3d697c185c21/jobs/6878

Not sure what to do TBH.

alysbrooks:
As a workaround, you can apply `^{:min-java-version "1.11"}` to the failing test (or I can). But it does seem to work sometimes.

test/shared/kaocha/integration_helpers.clj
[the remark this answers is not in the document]
alysbrooks: I think if you take the workaround I suggest below, you can just drop this.

plexus:
Thanks a lot @frenchy64 for this PR and @alysbrooks for shepherding it to completion, I really appreciate the back and forth here to make sure it's good. Hiding load errors is extremely frustrating for users, so I'm glad we get to weed out these issues.

I see this introduces a new way of doing kaocha black box testing, this is normally what we have Cucumber for, which has the benefit of doubling as documentation. But Ambrose I appreciate you already put a lot of time into this so I'm fine with having this go in as-is. Maybe put a comment on the test saying that we'd like to move this to cucumber at some point?

A CHANGELOG entry would also be good. Apart from that all looks good.

alysbrooks:
Thank you, @frenchy64!

I should have documented my reasoning for suggesting black box testing be done with Cucumber. If I remember correctly, my reasoning is that we already had integration tests for watch not part of Cucumber tests.

Issue #270, opened 2022-02-16
Error on Windows as of 1.62.993


skynet-gh:
I'm getting this error now when running kaocha on my project:

```
> clj -M:kaocha
java.nio.file.NotDirectoryException: tests.edn
 at sun.nio.fs.WindowsWatchService$Poller.implRegister (WindowsWatchService.java:379)
    sun.nio.fs.AbstractPoller.processRequests (AbstractPoller.java:266)
    sun.nio.fs.WindowsWatchService$Poller.run (WindowsWatchService.java:596)
    java.lang.Thread.run (Thread.java:833)
[watch] watching stopped.
```

[deps.edn alias](https://github.com/skynet-gh/skylobby/blob/master/tests.edn) and [tests.edn](https://github.com/skynet-gh/skylobby/blob/master/tests.edn)

Looks like the file watcher changes are causing it to think `tests.edn` should be a directory for some reason. Adding `:kaocha.watch/type :hawk` does not fix it.

oxalorg:
> :kaocha.watch/type :hawk does not fix it

Hmmmm! This is weird. Because using that flag it should have reverted back to the old watching mechanism. Maybe there has been an unintended change in this release? 🤔 

Thanks for reporting this issue I personally don't use windows so it might be hard to debug this! Any help would be appreciated.

At the risk of asking the obvious, does the immediate last release of Kaocha work fine for you?

skynet-gh:
> > :kaocha.watch/type :hawk does not fix it
> 
> Hmmmm! This is weird. Because using that flag it should have reverted back to the old watching mechanism. Maybe there has been an unintended change in this release? 🤔

Actually, I was wrong, I was putting `:kaocha.watch/type :hawk` in the test config and not top-level. Putting it top-level works!

> At the risk of asking the obvious, does the immediate last release of Kaocha work fine for you?

It does not but I'll use the workaround of hawk for now, thanks!

alysbrooks:
Looking at the Beholder repository, it looks like Beholder supports Windows, so I'm not sure why this failed. Based on the error, it could be a bug in the `WindowsWatchService`. In that case, it may be a version-specific issue.

alysbrooks:
So Beholder doesn't support file-level watching, which we try to do for `tests.edn`. Apparently the WindowWatchService notices, and the ones for Linux and macOS don't (and instead silently fail). There's a couple options:

[...]

PR #378, opened 2023-01-10
Fix Windows crash when using `--watch`

alysbrooks:
Not only does  watching individual files not work with Beholder, when using Beholder on Windows, it actually crashes. This PR disables watching `tests.edn` when using Beholder as as temporary fix, as proposed by @colin-p-hill in #270 .

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ## Fixed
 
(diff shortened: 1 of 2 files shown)


alysbrooks:
This is a safe change, so I'm going to go ahead and merge.

Issue #384, opened 2023-01-23
Kaocha crashes if test assertion runs after deftest function


vincentjames501:
We have a use-case where we have a fixture that does some final assertions after the test runs. Looks like a recent update to Kaocha changed some behavior here.

Here is a simple failure example:

```clojure
(ns kaocha-failure-test
  (:require [clojure.test :refer :all]))

(defn- post-assertion-fixture
  [f]
  (f)
  (is (= true true)))

(use-fixtures :each post-assertion-fixture)

(deftest my-test
  (is (= 2 (+ 1 1))))
```

```
$ ./kaocha --focus kaocha-failure-test
...
[(..Execution error (ClassCastException) at kaocha.testable/run (testable.clj:131).
class java.lang.Boolean cannot be cast to class clojure.lang.Associative (java.lang.Boolean is in module java.base of loader 'bootstrap'; clojure.lang.Associative is in unnamed module of loader 'app')

Full report at:
/var/folders/wn/cfyvfy_j60n7rzr9ndvgwtcm0000gp/T/clojure-8244869028492603707.edn
Error encountered performing task 'run' with profile(s): 'base,system,user,provided,dev'
Suppressed exit

$ cat /var/folders/wn/cfyvfy_j60n7rzr9ndvgwtcm0000gp/T/clojure-8244869028492603707.edn
{:clojure.main/message
 "Execution error (ClassCastException) at kaocha.testable/run (testable.clj:131).\nclass java.lang.Boolean cannot be cast to class clojure.lang.Associative (java.lang.Boolean is in module java.base of loader 'bootstrap'; clojure.lang.Associative is in unnamed module of loader 'app')\n",
 :clojure.main/triage
 {:clojure.error/class java.lang.ClassCastException,
  :clojure.error/line 131,
  :clojure.error/cause
  "class java.lang.Boolean cannot be cast to class clojure.lang.Associative (java.lang.Boolean is in module java.base of loader 'bootstrap'; clojure.lang.Associative is in unnamed module of loader 'app')",
  :clojure.error/symbol kaocha.testable/run,
  :clojure.error/source "testable.clj",
  :clojure.error/phase :execution},
 :clojure.main/trace
 {:via
  [{:type clojure.lang.Compiler$CompilerException,
    :message
    "Syntax error macroexpanding at (/private/var/folders/wn/cfyvfy_j60n7rzr9ndvgwtcm0000gp/T/form-init10811224181315925298.clj:1:126).",
    :data
[...]

PR #385, opened 2023-01-23
Fix :each fixture wrapping

NoahTheDuke:
Moving the `type/report-count` portion of `testable/-run :var` into `test-var` wrapped it in the `join-fixtures` calls, meaning that the `is` assertion was in the tail position and returned the result (a boolean) instead of returning the merged `testable` object.

To fix, I've pulled the `type/report-count` portion back into `testable/-run :var` and have added a top-level test that fail without this change.

Closes [#384](https://github.com/lambdaisland/kaocha/issues/384)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@
 
 - Configuration errors and circular dependencies are reported as warnings, 
     rather than causing the entirewatcher to crash.
+- Fix bug added in [#384](https://github.com/lambdaisland/kaocha/issues/384):
+    assertions in the tail position of an each fixture would return the result
+    of the assertion instead of the testable object with the merged report data.
+    (thanks [@NoahTheDuke](https://github.com/NoahTheDuke))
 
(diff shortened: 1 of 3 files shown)


NoahTheDuke:
The description is slightly more specific than necessary. The result of the tail call of test is returned, not just assertions.

alysbrooks:
I would have similar concerns as to the original PR (changing a fundamental part of the code that actually runs tests), but this partly reverts a previous change and we have tests for both. 

As I've thought about it more, I do wonder whether assertions in fixtures is really the best approach. Still, I don't think the inadvisability is a good reason to break compatibility, and I think the behavior this PR implements (returning the merged testable) makes more sense anyway.

NoahTheDuke:
Thanks so much. Sorry for the hassle. A real clear cut case of being too clever with my refactoring instead of touching the minimal lines to affect the desired change.

plexus:

```clj
```

PR #386, opened 2023-01-24
Make Kaocha compatible with Babashka

alysbrooks:
This requires Babashka 1.1.171 and disabling the `:notifier` plugin (for now).

In contrast to my experiment branch `alys/bb-experiment`, this PR doesn't use any conditionals. Instead, we create a new `classpath.bb` and a macro `if-babashka`

Why not use conditionals?
* Renaming files makes it harder to follow the history.
* Modifying the main source risks breaking Kaocha under regular Clojure (this disadvantage is shared by the macro).
* I think `.cljc` files may communicate that these namespaces run on ClojureScript.

Using conditionals does have the advantage of being something tried and true, compared to a new macro.

Remaining steps:
- [ ] Test on Kaocha's own test suite
- [x] Get notifier plugin working
- [x] Look into refactoring out Clojure/Babashka specifics.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Added
 
+* Kaocha is now compatible with Babashka. Running under Babashka is most useful
+    for validating your Clojure code runs correctly on Babashka, but it may also
+    be faster for some test suites because of reduced start up time.
+
 ## Fixed
 
 ## Changed
--- a/README.md
+++ b/README.md
@@ -204,6 +204,30 @@ If you wanted to turn off `fail-fast` temporarily, you could run `clojure
 -X:watch-test :fail-fast? false`
 
 
+### Babashka
+
+Kaocha is compatible with [Babashka](https://github.com/babashka/babashka). 
+
+You can create a `bb.edn` file:
+
+```clojure
+{:paths ["src" "test"]
+ :deps {lambdaisland/kaocha {:mvn/version "1.77.1236"}}}
+
+```
+
+Then you can create a binstub named `bin/kaocha-bb`:
+
+```shell
+
+#!/usr/bin/env bash
+bb -m kaocha.runner/-main  $@
+
+```
+
+If you exclusively want to run tests using Babashka, you can of course call it
+`bin/kaocha`.
+
 ### All tools
 
 By default, Kaocha assumes that:
--- a/deps.edn
+++ b/deps.edn
@@ -35,5 +35,6 @@
   :dev
   {:extra-paths ["dev"]
    :extra-deps  {djblue/portal {:mvn/version "RELEASE"}}}
-  :bb 
-  {:extra-deps {nubank/matcher-combinators    {:mvn/version "3.5.1"}}}}}
+
+  :bb
+  {:extra-deps {nubank/matcher-combinators {:mvn/version "3.5.1"}}}}}
--- a/deps.edn
+++ b/deps.edn
@@ -34,4 +34,6 @@
 
   :dev
   {:extra-paths ["dev"]
-   :extra-deps  {djblue/portal {:mvn/version "RELEASE"}}}}}
+   :extra-deps  {djblue/portal {:mvn/version "RELEASE"}}}
+  :bb 
+  {:extra-deps {nubank/matcher-combinators    {:mvn/version "3.5.1"}}}}}
--- a/doc/02_installing.md
+++ b/doc/02_installing.md
@@ -166,3 +166,27 @@ Kaocha will not be available in your REPL by default.
 ### Boot
 
 See [kaocha-boot](https://github.com/lambdaisland/kaocha-boot) for instructions.
+
+### Babashka
+
+Kaocha is compatible with Babashka. The main reason to run tests under Babashka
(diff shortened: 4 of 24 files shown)


plexus:
Looking great @alysbrooks , I riffed off this some more here: https://github.com/lambdaisland/kaocha/pull/387 . Not fully baked yet but we're clearly in downhilll territory now.

cc @borkdude @fdserr

borkdude:
@plexus @alysbrooks Great progress. If you want, I'd be happy contribute a github actions yml file where we execute tests in bb `latest`. If you find any strange things that should be solved in bb or have any questions, ping me.

alysbrooks:
@borkdude If it's something existing you can mostly just drop in that would be helpful. We might want to move it to CircleCI for consistency with the rest of our tests, so I don't want you putting  in a ton of work to adapt it.

borkdude:
@alysbrooks My thought is that adding a github actions runner for bb doesn't interfere as much with your existing tooling.

Also check out this comment where we did a similar thing for deep-diff2:

https://github.com/lambdaisland/deep-diff2/pull/36#issuecomment-1327240979

alysbrooks:
I think moving Arne's refactorings to `kaocha.platform.*` makes sense. I think these are inherently a lower layer and if we ever add support for other runtimes (possibly ones that don't even exist yet), this will help organize that work.

plexus:
Careful about making breaking changes, there might be plugins that rely on kaocha.classpath, I don't think we can move that.

borkdude:
@plexus Is kaocha.classpath intended as a public API? A quick glance on grep.app shows that no public repositories are using it yet: https://grep.app/search?q=kaocha.classpath. Nevertheless, may be good to err on the safe side.

plexus:
I think it would be valid for a plugin to rely on that API (same for other internal libraries like kaocha.testable or kaocha.load). We generally avoid renaming things for the sole reason that "it's nicer this way". It potentially disrupts the workflow of someone downstream, or causes churn for downstream maintainers, and so it's not worth the risk, even if the chance is quite small.

borkdude:
I agree, as long as it's clear what is public and private API: never change the public API.

alysbrooks:
Kaocha-cloverage actually does use `kaocha.classpath` (and that's actually why the tests are failing). Obviously we can change it, but I think it illustrates Arne's point that other libraries might use it.

borkdude:
It seems the test still works on the main branch, right? If so, then there must be a chance which makes this test fail... somewhere? :)

alysbrooks:
Slight tweak to some of our integration tests to prevent them from hanging on failure. Now they still fail, but do so right away, like you'd hope.

As I included in my last commit message, I don't know if there are some situations where the process would still be unspooling information, and blocking on slurp for a brief moment allows for capturing that information. I think missing out on that is acceptable since you are already in a failure state and hanging is much worse. Adding a short delay could mitigate this, however.

alysbrooks:
I nearly forgot. The change causing the test to fail is `kaocha.util/compiler-exception-file-and-line`. It slightly tweaks the output when tests fail, causing that integration test to no longer match. The exact change is that instead of just printing the file name, we now print the whole path.

alysbrooks:
Kaocha's own test suite is problematic, but it seems to run well on more typical test suites.

plexus:
I'm merging this, I think we've gone over this thoroughly enough at this point. Let's cut a release and see if anyone runs into issues.

plexus:
@alysbrooks what's this about? Our matcher-combinators version is pinned to 1.5.x, we never migrated to the newer versions, so this will certainly make our tests fail.

```clj
;; deps.edn
{,,,
 :aliases
 {,,,
  :bb
  {:extra-deps {nubank/matcher-combinators {:mvn/version "3.5.1"}}}}}
```

plexus:
Released in [v1.79.1270](https://github.com/lambdaisland/kaocha/releases/tag/v1.79.1270)

```clj
[lambdaisland/kaocha "1.79.1270"]                 ;; deps.edn
{lambdaisland/kaocha {:mvn/version "1.79.1270"}}  ;; project.clj
```

borkdude:
@plexus @alysbrooks Exciting! I think one thing might still be missing: some tests (at least some smoke tests) against `bb latest` so it's apparent when something is introduced that breaks bb compatibility. If you're open to a PR, I willing to to this work (later this week) like I did for some other libs recently, if it's ok to run a Github runner. If CircleCI is a must, then I'd rather defer that to your team since I'd have to figure that out more.

plexus:
A PR for that would be great @borkdude , and using GH actions is fine too. Thank you! I personally dislike Circle and GH actions in equal amounts so no particular preference ;)

plexus:
image

borkdude:
I'm also using both, but lambdaisland is doing way more advanced stuff with CircleCI than I have :)

plexus:
Yeah I once put a bunch of effort into creating these docker images and associated "orbs", so people would have a convenient way of setting up clojure/kaocha on circle ci , but in the end we never really promoted them so I guess it's mostly just us using them. It's also a mess of YAML templating which... does not spark joy.

PR #387, opened 2023-01-24
More babashka compat

plexus:
Generally try to move towards code that simply works on both platforms, when that's not possible factor out platform-specific code in platform-specific files (.bb vs .clj).

This way `if-babashka` isn't needed.

Note that normally we have a `.platform` namespace with all the platform specific bits, e.g. https://github.com/lambdaisland/uri/tree/main/src/lambdaisland/uri , but `kaocha.platform` was already taken, and I ended up introducing separate `kaocha.classpath` and `kaocha.systray` split namespaces. Maybe `kaocha.platform.systray` and `kaocha.platform.classpath` would be better.

`watch-load-error-test` seems to hang right now, probably due to the exception handling refactoring. I like how it simplifies things, but it does introduce a bit of change. I think relying on specific ex-data keys is a more useful contract than looking for a specific (nested) compiler error class.

### DIFF
--- a/deps.edn
+++ b/deps.edn
@@ -35,5 +35,6 @@
   :dev
   {:extra-paths ["dev"]
    :extra-deps  {djblue/portal {:mvn/version "RELEASE"}}}
-  :bb 
-  {:extra-deps {nubank/matcher-combinators    {:mvn/version "3.5.1"}}}}}
+
+  :bb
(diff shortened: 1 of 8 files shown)


alysbrooks:
This was kind of what I was hoping to do more of after I completed the main PR (itself a revision), but wasn't sure how. I think I'll merge your changes and iterate on them that way.

PR #391, opened 2023-01-30
Remove unnecessary requires

alysbrooks:
Added some comments about some edge cases.

Unnecessary `require`s aren't a major issue, but there are advantages to cleaning them up. This PR silences some warnings from `clj-kondo`, makes the namespaces easier to understand, and may avoid some circular imports in the future.


src/kaocha/runner.clj
```diff
@@ -15,7 +15,7 @@
             [kaocha.output :as output]
             [kaocha.plugin :as plugin]
             [kaocha.result :as result]
-            [kaocha.specs :as specs]
+            #_[kaocha.specs :as specs] ;; required for some reason?
```
borkdude, line 18: If you require namespaces for side effect, you can leave out the alias and clj-kondo won't report it as unused.

plexus: When namespaces are required for side-effects, the convention we use is to `(require '...)` them below the ns form.

borkdude: This also works (for clj-kondo)

alysbrooks:
This seems slightly more prone to breaking things not covered by tests, so I figure we should wait until #413 is merged.

alysbrooks:

```clj
```

PR #392, opened 2023-01-31
Load namespaces first required with :as-alias

### LINKED ISSUE #390: `:as-alias` in custom reporter namespace prevents target namespace tests from loading
In certain circumstances, a custom reporter which should be valid can cause a test to fail to load. A minimal repro can be found [here](https://github.com/colin-p-hill/kaocha-as-alias-repro).

It's a bit of a corner case. Here are the unusual-but-valid choices that led to this scenario:
1. Tests are defined inline in `src`.
2. The project hosts its own custom reporter.
3. The reporter namespace has an `as-alias` require of one of the project's other namespaces, the latter containing tests.

This results in Kaocha not running the tests in the aliased namespace. However, they appear in the test plan when printed. When running in `--watch` mode, these tests do not run initially, but they do run when the watcher refreshes the namespace that contains them.

To justify one of those choices: My use case for the `:as-alias` is to alter reporter behavior based on a namespaced metadata flag. For portability, I wanted to use a namespace which does not make specific reference to Kaocha. This meant it had to be a different namespace than the one that holds my custom reporter, since the reporter is of course coupled tightly to Kaocha – hence, `:as-alias`.

I haven't tested without criterion 1, but I suspect that it isn't strictly a necessary condition. If I had separate `src` and `test` directories, I would guess that an `:as-alias` pointing to a `test` namespace would result in skipping that namespace's tests – but I think this is probably a less reasonable thing to do, so I'm presenting a more realistic use case.

A workaround exists in simply using the full namespace in my references to the keyword, rather than an autoresolving alias. But this was very surprising behavior, and the root cause of the error is very much not obvious – it took about two hours to debug, which involved starting with my entire codebase and deleting parts until I was left with the minimal case. Hate for someone else to run into the same mystery.

alysbrooks:
Ensure libs are loaded (and don't just have an alias) before skipping them in our loading code in kaocha.ns. Fixes #390.

### DIFF
--- a/src/kaocha/ns.clj
+++ b/src/kaocha/ns.clj
@@ -9,7 +9,8 @@
             [kaocha.type :as type]))
 
 (defn required-ns [ns-name]
-  (when-not (find-ns ns-name)
+  (when-not (and (find-ns ns-name)
+                 (contains? (loaded-libs) (symbol ns-name)))
     (require ns-name))
   (try
     (the-ns ns-name)


alysbrooks:
We could possibly skip the test and just use `require`, since `require` doesn't reloaded already existing namespaces, or just use the check I added (with `loaded-ns`). But I don't know the rationale for not just using `require` and this seems the best way to get the fix out.

alysbrooks:
Note to self: I should add a test before merging this.

alysbrooks:
I don't want to delay releasing this fix anymore, so I'm going to go ahead and merge it without tests.

PR #393, opened 2023-02-06
332 plugin kaocha plugin profiling does not run when watch is used

fdserr:
Wrap the callsite to api/run in the watcher with the post-summary hook, exactly as in the runner.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ## Fixed
 
+- The watcher now prints the output from plugins, eg. `bin/kaocha --watch --plugin kaocha.plugin/profiling` 
+
 ## Changed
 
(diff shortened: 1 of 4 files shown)


plexus:
This looks good, but it changes some of the watch output, so some of the tests need updating. And it has conflicts with `main` at this point. @alysbrooks do you think you can wrap this up? I don't think we can still count on François.

alysbrooks:
Yes, I can fix it up,

alysbrooks:
Released in [v1.83.1314](https://github.com/lambdaisland/kaocha/releases/tag/v1.83.1314)

```clj
[lambdaisland/kaocha "1.83.1314"]                 ;; deps.edn
{lambdaisland/kaocha {:mvn/version "1.83.1314"}}  ;; project.clj
```

PR #396, opened 2023-02-18
Fix Cljdocs (again)

alysbrooks:
I realized my quick fix won't work because the other namespaces aren't available in ClojureScript either. So this change will prevent Cljdocs from trying to analyze Kaocha as a ClojureScript project.

### DIFF
--- a/doc/cljdoc.edn
+++ b/doc/cljdoc.edn
@@ -1,0 +1,1 @@
+{:cljdoc/languages ["clj"]}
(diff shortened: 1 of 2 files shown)


alysbrooks:
Released in [v1.78.1249](https://github.com/lambdaisland/kaocha/releases/tag/v1.78.1249)

```clj
[lambdaisland/kaocha "1.78.1249"]                 ;; deps.edn
{lambdaisland/kaocha {:mvn/version "1.78.1249"}}  ;; project.clj
```

PR #398, opened 2023-02-22
Update Beholder to improve support for certain editors.

alysbrooks:
Beholder released a fix for certain editors that create a new file and then delete instead of modify. I believe Vim and Neovim do this when `noswapfile` and either `backupcopy=no` or `backupcopy=auto` are set, but other editors probably do it too.

Sometimes these creation events would be dropped because the file exists already. I haven't noticed this problem when using --watch with Neovim, but it may depend on timing and thus may be highly platform-, version- or setup-specific. Also, I consider this a safe change to make without reproducing the initial issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Added
 
 ## Fixed
+* Fix issue with `--watch` and Neovim by bumping Beholder to `1.0.2`
 
 ## Changed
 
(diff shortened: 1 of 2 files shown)


alysbrooks:
Released in [v1.78.1249](https://github.com/lambdaisland/kaocha/releases/tag/v1.78.1249)

```clj
[lambdaisland/kaocha "1.78.1249"]                 ;; deps.edn
{lambdaisland/kaocha {:mvn/version "1.78.1249"}}  ;; project.clj
```

PR #400, opened 2023-03-02
Run bb smoke test on CI

borkdude:


borkdude:
@alysbrooks @plexus This PR adds a smoke test for bb on CI.

Notes:
- I included bb tests in `test/bb`
- I added an entry in `tests.edn` for bb, but maybe I should use a different config file for bb tests? Does kaocha support providing a config on the command line without creating a file?
- I added a return value in the notifier plugin mock for bb as to not crash when it is used in bb

borkdude:
I see the tests are failing because of unexpected keys in the `tests.edn` file:

```
:kaocha.testable/desc "bb (clojure.test)"
```

alysbrooks:
> * Does kaocha support providing a config on the command line without creating a file?

You can specify a particular `tests.edn` using `--config-file`. There isn't one for `test-paths`, at least not in our regular runner. I think you can use the exec-fn to do that, technically: https://cljdoc.org/d/lambdaisland/kaocha/1.80.1274/doc/readme#clojure-cli-toolsdeps-exec-fn-alternative.

borkdude:
No option to pass config as edn without writing a file? If not, then I'll add another `bb-tests.edn` file.

alysbrooks:
Might be good to add, but yeah, there's not a way to do it right now. You can include another EDN file since we have [Aero](https://github.com/juxt/aero) support.

borkdude:
@alysbrooks OK, this should settle it then.

alysbrooks:
Thanks, @borkdude! 

Banner reading THANK YOU in blue an moving up and down

borkdude:
Woohoo!

plexus:
Released in [v1.81.1290](https://github.com/lambdaisland/kaocha/releases/tag/v1.81.1290)

```clj
[lambdaisland/kaocha "1.81.1290"]                 ;; deps.edn
{lambdaisland/kaocha {:mvn/version "1.81.1290"}}  ;; project.clj
```

borkdude:
Dank!

Issue #162, opened 2020-10-09
Ability to disable ::zero-assertions as fail condition


jaybarra:
I would like the ability to disable via configuration the zero-assertions condition as a failing condition. There are several tests where the `is` has been moved to a separate namespace function call which are being falsely flagged as having no conditions.

plexus:
Generally I would recommend writing your test in such a way that there's always an assertion. That said making this configurable doesn't hurt, PR welcome.

ourkwest:
We are running some end-to-end UI tests and prefer to allow our UI driver library to throw assertions when things aren't there.
This gives us fail fast (which we want in this scenario) for free, but means that most tests do not need a clojure.test assertion.

Would a PR still be welcome?

(At the moment I'm hacking the kaocha.hierarchy, but I'm not proud of what I've done.)

plexus:
Yes, a PR would still be most welcome. Thanks!

ourkwest:
Having looked at the code it seems that kaocha's hook mechanism already makes this reasonably straightforward. Due to shortness of time I've used that instead. If anyone else comes here looking to disable the fail on zero assertions behaviour here's what I did:

```
(ns my.kaocha.hooks
  (:require
    [kaocha.hierarchy :as hierarchy]))

(hierarchy/derive! ::ignore :kaocha/known-key)

(defn defuse-zero-assertions [event]
  (if (= (:type event) :kaocha.type.var/zero-assertions)
    (assoc event :type ::ignore)
    event))
```

I put that code in a namespace and configured it as a hook in my tests.edn file:

```
#kaocha/v1
{:plugins [:kaocha.plugin/hooks]
 :kaocha.hooks/pre-test [my.kaocha.hooks/defuse-zero-assertions]}
```

plexus:
I'm surprised that actually works. Are you sure you used a `pre-test` hook and not a `pre-report` hook?

humorless:
> I'm surprised that actually works. Are you sure you used a `pre-test` hook and not a `pre-report` hook?

I re-run the example above, and I found: 

(1) If I run the example precisely, I will get 
```
Execution error (ArityException) at kaocha.plugin.hooks/hooks-pre-test-hook$fn (hooks.clj:95).
Wrong number of args (2) passed to: my.kaocha.hooks/defuse-zero-assertions
```

(2) If I run the example with changing `:kaocha.hooks/pre-test` to `:kaocha.hooks/pre-report`, then it works. 

[...]

PR #404, opened 2023-03-16
add config for zero assertions switcher in config and cli args

### LINKED ISSUE #211: Interpret the "0 test" situation as a failure

humorless:
## This is for #162 

kaocha's default behavior will let the tests fail if there are no `is` in assertion. 
```
laurencechen qqq $ ./bin/kaocha  
[(F)]
Randomized with --seed 1954406926

FAIL in qqq-test/example-test (qqq_test.clj:5)
Test ran without assertions. Did you forget an (is ...)?
1 tests, 1 assertions, 1 failures.
```

With this MR, we now support a new config/cli args to switch off the `zero-assertions`

Example of `tests.edn`

```
#kaocha/v1
{:tests [{:id          :unit
          :test-paths  ["test" "src"]
          :ns-patterns [".*"]}]
:warnings {:zero-assertions :silent}
}
```

## This MR can also fix #211 
Example of `tests.edn`

```
#kaocha/v1
{:tests [{:id          :unit
          :test-paths  ["test" "src"]
          :ns-patterns [".*"]}]
:warnings {:zero-tests :error}
}
```
;; => Results
截圖 2023-04-21 下午4 55 36

### DIFF
--- a/src/kaocha/hierarchy.clj
+++ b/src/kaocha/hierarchy.clj
@@ -8,6 +8,11 @@
   [tag parent]
   (alter-var-root #'hierarchy derive tag parent))
 
+(defn underive!
+  "Add a parent/child relationship to kaocha's keyword hierarchy."
+  [tag parent]
+  (alter-var-root #'hierarchy underive tag parent))
+
 (derive! :fail :kaocha/fail-type)
 (derive! :error :kaocha/fail-type)
 
--- a/doc/03_configuration.md
+++ b/doc/03_configuration.md
@@ -20,6 +20,7 @@ Here's an example test configuration with a single test suite:
                                        :kaocha/source-paths  ["src"]
                                        :kaocha/test-paths    ["test/unit"]}]
  :kaocha/fail-fast?                  false
+ :kaocha/mute-zero-assertion?        true
  :kaocha/color?                      true
  :kaocha/reporter                    [kaocha.report/dots]
  :kaocha/plugins                     [:kaocha.plugin/randomize
--- a/doc/03_configuration.md
+++ b/doc/03_configuration.md
@@ -20,7 +20,6 @@ Here's an example test configuration with a single test suite:
                                        :kaocha/source-paths  ["src"]
                                        :kaocha/test-paths    ["test/unit"]}]
  :kaocha/fail-fast?                  false
- :kaocha/mute-zero-assertion?        true
  :kaocha/color?                      true
  :kaocha/reporter                    [kaocha.report/dots]
  :kaocha/plugins                     [:kaocha.plugin/randomize
--- a/doc/clojure_test/assertions.md
+++ b/doc/clojure_test/assertions.md
@@ -32,6 +32,12 @@ FAIL in sample-test/my-test (sample_test.clj:4)
 Test ran without assertions. Did you forget an (is ...)?
 ```
 
+### Missing assertions are configurable
+
+There are two ways to disable this detection:
+
+* Enable `:mute-zero-assertion?` at `tests.edn`.
+* Use `--mute-zero-assertion` in command line argument.
 
 
 ## Detecting single argument `=`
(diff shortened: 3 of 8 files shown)


src/kaocha/hierarchy.clj
```diff
@@ -8,6 +8,11 @@
   [tag parent]
   (alter-var-root #'hierarchy derive tag parent))
 
+(defn underive!
+  "Add a parent/child relationship to kaocha's keyword hierarchy."
```
alysbrooks, line 12: This removes a parent/child relationship, right?

alysbrooks:
Since we might want to make other warnings configurable, I think we should either:
* Accept a list of warnings to mute like `:mute-warnings [:zero-assertions]`.
* Allow for both increasing the warning level (to a failing error) and muting it:  `:warnings {:zero-assertions :silent :unwrapped-values-is :error}`. (`:unwrappd-values-is` is a hypothetical other warning we could cover.)

I didn't closely review the rest of the code (although it looks good on first glance) since I think it will have to change if you take one of my suggestions.

src/kaocha/hierarchy.clj
humorless: yes

humorless:
> #162

@plexus Arne, can you also provide your ideas on my change? Alys provided me a **configuration option/ arguments design** seems more elegant, but I am not sure should we do it now.

plexus:
I agree with Alys, we have a number of warnings that could be configurable. I would rather not introduce a new top level key for each. Instead we should introduce a pattern that people can easily extend when they want to make other warnings configurable.

alysbrooks:
I'm leaning toward the latter option (the `:warnings` map), for what it's worth. That would let us support warnings that are turned off by default (perhaps because they're alpha) and allow for a stricter configuration on CI.

humorless:
@alysbrooks @plexus 

I have implemented the config argument as you suggested. On the other hand, in this case, do I still need to implement `warnings` in command-line-arguments? And if so, how to do it? Any suggestions?

alysbrooks:
@humorless Yes, that's tricky. I'm not sure there's a great way to do key-value command line arguments. (I found some discussion [here](https://stackoverflow.com/questions/18115674/how-to-format-command-line-argument-key-value-pairs)). Maybe we could use an approach similar to  `-Jkey=value` like Clojure CLI? (I guess we could use W for warning? Or E for error?) Although maybe `--warning name=level` would be better?

Usually I'd expect people would modify it in `tests.edn`, but I suppose you could have a situation where you want to disable it on the command line because you're getting false positives as you work on a test? Perhaps we could save the flag for a later PR?

src/kaocha/hierarchy.clj
alysbrooks: Then I'd suggest:

```suggestion
  "Remove a parent/child relationship from Kaocha's keyword hierarchy."
```

alysbrooks:
> By the way, I somewhat prefer to delay the command line argument design for later PR. I really think that is there really a need to control kaocha to this level with command line arguments?

Yes, that makes sense to me.

src/kaocha/hierarchy.clj
alysbrooks: Are you planning to make this change?

alysbrooks:
A possible refactoring would be a function that outputs using the correct level based on the warning. For example, `(notify warnings :some-key "message")` so we don't have to repeat the `(if ((:some-warning warnings) :warning) ,,,)` code multiple times. But maybe we should first look at the rest of the code base?

PR #406, opened 2023-03-27
add windows binstub

rgkirch:


alysbrooks:
Thanks, @rgkirch! This is definitely a gap in our documentation.

What's the advantage of creating a `!/usr/bin/env sh` script that calls PowerShell over a `.ps1` script?

rgkirch:
So, I use msys2 and the incantation that I need is `powershell -command clojure -M:test "$@"`. If I'm in powershell and I invoke `.\bin\kaocha.ps1` then I need `powershell -command clojure -M:test $args`. Powershell doesn't handle the terminal color things and gives me some not pretty output:
```
←[33mWARNING: ←[mDid not load a configuration file and using the defaults.
This is fine for experimenting, but for long-term use, we recommend creating a configuration file to avoid changes in behavior between releases.
To create a configuration file using the current defaults and configuration file location, create a file named tests.edn that contains '#kaocha/v1 {}'.
Reflection warning, meander/util/epsilon.cljc:758:24 - reference to field val can't be resolved.
Success!
[(←[31mF←[m.←[31mF←[m.←[31mF←[m)]
Randomized with --seed 955536687

←[31mFAIL←[m in transform-test/extra-stuff-test (transform_test.clj:7)
```
My msys2 terminal is fine though. I don't actually use powershell for anything but clojure installs as a powershell module (whatever that is).

rgkirch:
No, I don't think you get sh by default.

plexus:
Right, in that case it's probably better to recommend that people add a separate `kaocha.bat` to their projects if they want to support windows users as well.

plexus:
Thank you for your ocntribution to Kaocha!

plexus:
Released in [v1.82.1306](https://github.com/lambdaisland/kaocha/releases/tag/v1.82.1306)

```clj
[lambdaisland/kaocha "1.82.1306"]                 ;; deps.edn
{lambdaisland/kaocha {:mvn/version "1.82.1306"}}  ;; project.clj
```

PR #407, opened 2023-03-30
Feature: exec-args support for print-config & print-test-plan

### LINKED ISSUE #401: print-config cannot be enabled from exec-fn
Seemingly related to #312, I can't seem to find a way to enable `print-config` from an `exec-fn` -X invocation. And [this line](https://github.com/lambdaisland/kaocha/blob/8e9cf5f17673c5b58784eefe4aac1c5ea62ab127/src/kaocha/runner.clj#L99) being the only hit for `print-config` when searching the repo makes me think it likely isn't currently possible.

cap10morgan:
This adds support for the following exec-args when using the Clojure CLI exec-fn / -X invocation method:

- `:print-config true`
- `:print-test-plan true`

Fixes #401

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Added
 
+- Expose `:print-config true` and `:print-test-plan true` to Clojure CLI -X
+  invocations.
+
 ## Fixed
 
 ## Changed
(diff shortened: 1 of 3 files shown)


plexus:
Thank you!

plexus:
Released in [v1.82.1306](https://github.com/lambdaisland/kaocha/releases/tag/v1.82.1306)

```clj
[lambdaisland/kaocha "1.82.1306"]                 ;; deps.edn
{lambdaisland/kaocha {:mvn/version "1.82.1306"}}  ;; project.clj
```

PR #411, opened 2023-05-06
Redo watch test as a vector of lines instead of a string.

alysbrooks:
This test is difficult to work with as-is because it's one big string, and deep-diff2 (currently) shows only that two strings differ, and not the substrings in particular that cause the mismatch.


joannecheng:
Closing for now to prioritize other tickets.

PR #415, opened 2023-05-12
Clean up output of :print-invocations

plexus:
When invoking with no extra arguments the output of `:print-invocations` would have a stray (double) space. e.g.

```
bin/kaocha  --focus 'kaocha.type.ns-test/run-test-parallel'
-----------^
```

Get rid of that.


alysbrooks:

```clj
```

Issue #416, opened 2023-05-17
Sending SIGTERM to `--watch` throws ClassCastException "java.io.OutputStreamWriter cannot be cast to java.io.PrintStream"


grzm:
I generally run kaocha with `--watch`, do some development, and when I’m done, type CTRL-C (sending SIGINT, I believe). Starting with release v[1.79.1270](https://github.com/lambdaisland/kaocha/releases/tag/v1.79.1270) I’ve started to see the following stacktrace when I exit:

```
^CException in exit (SIGINT/ShutDown) handler
#error {
 :cause "java.io.OutputStreamWriter cannot be cast to java.io.PrintStream"
 :via
 [{:type java.lang.ClassCastException
   :message "java.io.OutputStreamWriter cannot be cast to java.io.PrintStream"
   :at [kaocha.api$run$fn__3265$fn__3269$fn__3270$on_exit__3271 invoke "api.clj" 125]}]
 :trace
 [[kaocha.api$run$fn__3265$fn__3269$fn__3270$on_exit__3271 invoke "api.clj" 125]
  [kaocha.api$run$fn__3265$fn__3269$fn__3270$fn__3273$fn__3274$fn__3275 invoke "api.clj" 141]
  [kaocha.api$run$fn__3265$fn__3269$fn__3270$fn__3273$fn__3274 invoke "api.clj" 140]
  [kaocha.api.proxy$java.lang.Object$SignalHandler$d8c00ec7 handle nil -1]
  [sun.misc.Signal$1 run "Signal.java" 212]
  [java.lang.Thread run "Thread.java" 750]]}
  ```
  
This doesn’t occur with the release just prior (v[1.78.1249](https://github.com/lambdaisland/kaocha/releases/tag/v1.78.1249)). It looks like it might be due to this commit? https://github.com/lambdaisland/kaocha/commit/09eccdbfb3eef69f1f35bfe2b421187acf78b6f7

alysbrooks:
@grzm You were only having issues with Ctrl-C (SIGINT), right? The original title mentioned SIGTERM.

I actually tried out SIGTERM, and it just immediately ends. The JVM may actually handle that signal itself.

plexus:
Seems this might be fixed in #417 ?

PR #417, opened 2023-05-18
fix the issue from kaocha #416: casting *out* to System/out

humorless:
# Problem

From #416, there is a bug raised. The bug is because that Clojure's `*out*` is not totally equivalent to java's `System/out`. Clojure's `*out*` is equivalent to System/out, wrapped in an OutputStreamWriter. Therefore, it caused a Casting error. 

# Solution

After referencing how output capture is implemented in `src/kaocha/plugin/capture_output.cljc`, although I think there is already a `finally` section, which should restore the java's `System/out` and `System/err`, I tend to believe that Arne's comment is based on his real world experiments. 

```
;; Force reset printing to stdout, since we
;; don't know where in the process we've
;; been interrupted, output capturing may
;; still be in effect.
```

### DIFF
--- a/src/kaocha/api.clj
+++ b/src/kaocha/api.clj
@@ -17,7 +17,8 @@
 ;; Prevent clj-refactor from "cleaning" these from the ns form
 (require 'kaocha.monkey-patch)
 
-(def orig-out *out*)
+(def orig-out System/out)
+(def orig-err System/err)
 
 (def ^:dynamic *active?*
   "Is Kaocha currently active? i.e. loading or runnning tests."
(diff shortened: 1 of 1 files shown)


alysbrooks:
Thanks, @humorless! Looks good.

I think adding your PR message (or a shorter version) to a comment would be good. If I remember correctly when I made the change, I looked around for comments and past commit messages but didn't find anything explaining why it was the way it was.

humorless:
> Thanks, @humorless! Looks good.
> 
> I think adding your PR message (or a shorter version) to a comment would be good. If I remember correctly when I made the change, I looked around for comments and past commit messages but didn't find anything explaining why it was the way it was.

Thanks for you suggestion. I have added the necessary comments to explain the bug.

alysbrooks:

```clj
```

PR #425, opened 2023-08-29
Fix watching when paths don't exist

simon-katz:
This fixes a problem where watching silently fails to happen when paths don't exist.

Tests (with `bin/kaocha`) are still passing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Added
 
+- Fix watching when paths do not exist.
+
 ## Fixed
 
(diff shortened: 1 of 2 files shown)


plexus:
Released in [v1.86.1355](https://github.com/lambdaisland/kaocha/releases/tag/v1.86.1355)

```clj
[lambdaisland/kaocha "1.86.1355"]                 ;; deps.edn
{lambdaisland/kaocha {:mvn/version "1.86.1355"}}  ;; project.clj
```

Issue #388, opened 2023-01-24
Fixtures not ending in test result do not work with kaocha v1.73.1175 onwards


Wegi:
Description:

When tests use fixtures like the following:

```clojure
(defn init-execute-teardown-fixture
  [f]
  (init-db-data)
  (f)
  (clean-db))
```

Then kaocha tries to cast the result of the `clean-db` call and not the test. E.g. if it is a Boolean, you get the exception:
`class java.lang.Boolean cannot be cast to class clojure.lang.Associative (java.lang.Boolean is in module java.base of loader 'bootstrap'; clojure.lang.Associative is in unnamed module of loader 'app')`

The fixture pattern above is the usual fixture pattern and even used in multiple examples in the official documentation.

A temporary fix is to return the test result from the fixture, like so:
```clojure
(defn init-execute-teardown-fixture
  [f]
  (init-db-data)
  (let [test-result (f)]
    (clean-db)
    test-result))
```

I can provide more example data, but this should be easy to reproduce.

Thanks!

plexus:
Can you try 

```
[lambdaisland/kaocha "1.76.1230"]
{lambdaisland/kaocha {:mvn/version "1.76.1230"}}
```

please? I believe we merged a fix for this recently.

Wegi:
Worked, flawlessly. Thanks for the work. 👍

NoahTheDuke:
We've come across an issue related to this: If you run the test in a fixture twice and only the first instance fails, then the fail is printed but the returned result says that there are no failures.

Given:
```clojure
(def ^:dynamic *val* 0)

(defn doubled-fixture [f]
  (binding [*val* 1]
    (f))
  (binding [*val* 2]
    (f)))

(use-fixtures :once doubled-fixture)

(deftest example-fail-test
  (is (= *val* 2)))
```

Output:
```
$ clojure -M:test
[(F.)]

FAIL in noahtheduke.example.repo-test/example-fail-test (repo_test.clj:16)
Expected:
  1
Actual:
  -1 +2
1 tests, 1 assertions, 0 failures.

$ echo $?
0
```

PR #432, opened 2024-03-22
Once Fixtures calling f multiple times

NoahTheDuke:
`:once` fixtures that call the provided `f` multiple times can swallow failures if the last call passes. Instead of swapping out the results of the `run-testables` call each time (which discards the existing failures), we should conj all calls together and return that as the completed result.

Related to #388

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@
 
 ## Fixed
 
+- Track all results when `:once` fixtures call `f` multiple times.
+  (thanks [@NoahTheDuke](https://github.com/NoahTheDuke))
+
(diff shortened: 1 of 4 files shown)


plexus:
Thanks @NoahTheDuke , really appreciate the structured approach you're taking here. Is this good to be merged?

plexus:
Oh and could you please add a change log entry? Thanks!

NoahTheDuke:
Excepting the change log entry, I believe it is.

plexus:
Much appreciated!

plexus:
This has been released in 1.88.1376.

Issue #440, opened 2024-08-12
defmethod watch! :beholder + wait-and-rescan! currently broken


ivarref:
Hi, and thanks for a great and useful project!

I noticed that adding `:kaocha.watch/ignore ["*.tmp"]` in `tests.edn` did not work as intended. As far as I can see there are two reasons for that:

`*.tmp` will only match files in the current directory.

`beholder` returns fully qualified names (on both mac and linux).

thus:

```clojure
(glob? (.toPath (java.io.File. "/foo/bar/demo.tmp")) ["*.tmp"])
=> nil
```

I suggest to update the documentation to give `**.tmp` as an example. That will match `/foo/bar/demo.tmp` as well.

The second reason is that `beholder` will return `java.nio.file.Path` and not `java.io.File`which `wait-and-rescan!` expects. This will lead to a NullPointerException.

I have written a pull request that fixes this.

Thanks again for a great project!
Kind regards
Ivar

plexus:
Thanks! This might be because we switched at some point from another watcher lib to beholder. Glad it's fixed!

PR #441, opened 2024-08-12
Fix beholder watch functionality. Improve docs.

ivarref:
Fixes https://github.com/lambdaisland/kaocha/issues/440

Thanks and kind regards

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ## Fixed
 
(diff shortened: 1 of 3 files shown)


plexus:
Leaving this open until I've had a chance to have a look at why CI fails. I'm assuming it's not actually an issue with this PR.

ivarref:
Thanks for the fast reply!

I made another commit that removes reflection warnings from `watch.clj`. Let me know if you'd like it squashed.

Regards.

PR #450, opened 2024-12-22
Resolve reflective calls in profiling plugin

frenchy64:


alysbrooks:
Looks good, but (assuming you're still interested in finishing this PR) can you remove the line setting `*warn-reflection*`?

PR #453, opened 2024-12-22
Make `kaocha.jit/jit` thread-safe

frenchy64:
Ensures we don't race with other requiring threads.

Note this is now safer than `requiring-resolve` (as of Clojure 1.12), which has a race condition making it prone to returning a partially initialized var.


alysbrooks:
The warning isn't a big deal either way, but I did look into the popularity of older Clojure versions, and 1.10 was already being phased out in 2022: https://clojure.org/news/2024/12/02/state-of-clojure-2024.

PR #463, opened 2025-08-28
Update pronunciation link on README.md

socksy:
Unfortunately, seems like linedict is no more, and the thing that replaced it doesn't seem to have pronunciation for 考察 (although it seems to be very happy to give five different pronunciations for the English translation "nose around"). 

Hanzii.net appears to be maintained, and has a nicer pronunciation (imo) than Forvo's Mandarin one. And Pleco is unfortunately still mobile only.


joannecheng:
thank you!

